PR #1, opened 2022-03-09
reviewing + format + easy rework

gsebil08:


lib/client.ml
```diff
@@ -1,17 +1,16 @@
 open Lwt_unix
-open Util
```
gsebil08, line 2: Feel free to change the `.ocamlformat` directly in this branch. I just copy/past the one from Deku for homogeneity.

lib/util.ml
```diff
@@ -4,17 +4,24 @@ module Encoding = struct
   open Bin_prot
   open Bin_prot.Utils
 
+  (* Bin_prot fixed size header. *)
+  let bin_prot_buffer_size = 8
```
gsebil08, line 8: I extracted these values to constantes, to avoid partial modifying.

lib/client.ml
```diff
@@ -1,40 +1,44 @@
...
+let peer_from (client : 'a t) =
+  Peer.{ address = client.address; port = client.port }
 
 let send_to client payload peer =
   let len = Bytes.length payload in
   let addr = Peer.to_sockaddr peer in
   let%lwt _ = sendto !client.socket payload 0 len [] addr in
   Lwt.return ()
 
+let send_to_peers client payload peers =
```
gsebil08, line 21: In order to prepare the gossip, I just created this function, to send the same message to several peers.

.ocamlformat
```diff
@@ -0,0 +1,23 @@
+version = 0.21.0
```
gsebil08, line 1: Feel free to change the .ocamlformat directly in this branch. I just copy/past the one from Deku for homogeneity.

lib/peer.ml
```diff
@@ -1,13 +1,26 @@
-type t =
-  { address: string
-  ; port: int
-  }
+type state =
```
gsebil08, line 1: This is the state a peer can have following `SWIM` protocol.

test/pollinate_tests.ml
```diff
@@ -5,67 +5,122 @@ module Client_tests = struct
   open Pollinate
 
   type request =
+    | Ping
```
gsebil08, line 8: Just the beginning of the `ping pong` for the ACK system.

test/pollinate_tests.ml
```diff
@@ -5,67 +5,122 @@ module Client_tests = struct
...
+    let res_a = Util.Encoding.unpack bin_read_response res_a in
+
+    let get = Util.Encoding.pack bin_writer_request Get in
+
+    let%lwt () = Client.send_to client_a get peer_b in
+    let%lwt status_of_b, _ = Client.recv_next client_a in
+
+    let status_of_b = Util.Encoding.unpack bin_read_response status_of_b in
+
+    let res_a, status_of_b =
```
gsebil08, line 81: This test is failing, I don't know what I miss, but I expect `status_of_b` to be `something` (the inserted value provided line 67).
Would you mind help me to figure our what I am missing?

rosalogia:
Made some  minor changes to main and rebased. Going to make a few branches off of this one to try some solutions to the problem I mentioned in my last review comment.

gsebil08:
>Going to make a few branches off of this one

I can remove the non-working test, so we can merge this one. What do you think?

rosalogia (changes requested):

lib/client.ml
gsebil08: Ok for the new naming, clearly better!
I didn't know `eta conversion` was present in OCaml! That's great!

test/pollinate_tests.ml
```diff
@@ -5,67 +5,122 @@ module Client_tests = struct
...
+      | List l1, List l2 -> (List.hd l1, List.hd l2)
       | _ -> failwith "Incorrect response" in
 
     Lwt.return (res_a, res_b)
+  let test_insert () =
+    let _ = Client.peer_from !client_a in
+
+    let peer_b = Client.peer_from !client_b in
+
+    let req = Util.Encoding.pack bin_writer_request @@ Insert "something" in
```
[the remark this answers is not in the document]
gsebil08: >I'm sure you're used to using $ all the time in Haskell 😄

Touché! Changed for `()` almost everywhere.

test/pollinate_tests.ml
```diff
@@ -5,67 +5,122 @@ module Client_tests = struct
...
+      | Ping -> Pong
+      | Get -> List !state
       | Insert s ->
         state := s :: !state;
         Success "Successfully added value to state" in
     Util.Encoding.pack bin_writer_response response
 
-    (* Initializes two peers and has each one request the state
-    of the other, returning the first element in the response of each *)
+  let client_a =
```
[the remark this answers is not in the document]
gsebil08: Yup, I knew it was not perfect, but it is much more simpler for testing purpose. Let's create an issue to track this, and maybe solve it in the future if needed.

lib/util.mli
```diff
@@ -1,24 +1,33 @@
 (** Defines utilities for working with payloads
 that are or need to be serialized via Bin_prot *)
 module Encoding : sig
+  val bin_prot_buffer : bytes
```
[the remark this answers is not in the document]
gsebil08: Seen in DM, it comes from `.ocamlformat` would be glad to figure out what setting I can use to fix this.

lib/util.ml
```diff
@@ -4,17 +4,24 @@ module Encoding = struct
...
+
+  let bin_prot_buffer = Bytes.create bin_prot_buffer_size
+
+  (* Position to start reading a datagram *)
+  let read_from_first_byte = 0
```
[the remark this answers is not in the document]
gsebil08: Totally agreed, you got my point on this badly naming 😉 .
Changed to your proposal which is good!

But I still have the two `0` within `client.recv_next`, should I also create a function? Or go for a constant like `let zero = 0`? (or any better naming, it looks like you have more skills that me on naming 🤣)

lib/util.ml
```diff
@@ -4,17 +4,24 @@ module Encoding = struct
   open Bin_prot
   open Bin_prot.Utils
 
+  (* Bin_prot fixed size header. *)
+  let bin_prot_buffer_size = 8
+
+  let bin_prot_buffer = Bytes.create bin_prot_buffer_size
```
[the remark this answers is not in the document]
gsebil08: >Bytes are mutable

😱 I didn't know that (but it's the first sentence of the documentation 😅)
I got rid of this and switch back to your proposal, which is better!

lib/client.ml
```diff
@@ -1,40 +1,49 @@
...
   let%lwt _ = sendto !client.socket payload 0 len [] addr in
   Lwt.return ()
 
+let naive_broadcast client payload peers =
+  let _ = List.map (send_to client payload) peers in
+  Lwt.return ()
+
 let recv_next client =
   let open Lwt_unix in
-  (* Peek at the first 8 bytes of the incoming datagram
```
rosalogia, line 24: Why don't we keep this comment or something to the same effect to explain the MSG_PEEK line

lib/client.ml
```diff
@@ -1,40 +1,49 @@
...
-  to read the Bin_prot size header. *)
-  let size_buffer = Bytes.create 8 in
+  let open Util.Encoding in
+  let buffer_size = Bytes.create size_header_length in
```
rosalogia, line 34: Let's call it size_buffer, otherwise it seems like this binding is a number rather than a buffer.

rosalogia: Or, to be consistent with msg_buff, we can call it size_buff

lib/client.ml
```diff
@@ -1,40 +1,49 @@
...
+let naive_broadcast client payload peers =
+  let _ = List.map (send_to client payload) peers in
+  Lwt.return ()
+
 let recv_next client =
   let open Lwt_unix in
-  (* Peek at the first 8 bytes of the incoming datagram
-  to read the Bin_prot size header. *)
-  let size_buffer = Bytes.create 8 in
+  let open Util.Encoding in
```
rosalogia, line 33: In my opinion (which is not necessarily the rule), it's nice to open up till the last module (e.g. when using functions from Util.Encoding, only do `let open Util in ...`) so that it's obvious where we call functions that weren't defined in this file. In other words, any function from `Util.Encoding` have `Encoding.` before them, so it's obvious they're not from this module. Otherwise, you have to check via the editor features which are not present in GitHub for example. What do you think?

lib/client.ml
```diff
@@ -43,10 +52,11 @@ let serve client msg_handler =
   Lwt.async server
 
 let init ~state ~msg_handler (address, port) =
-  let%lwt socket = Net.create_socket port in
+  let open Util.Net in
```
rosalogia, line 55: See previous comment about module opening, but know that it's not a must for me.

rosalogia (changes requested):
Sorry, just a few more things, and they are quite small :)

lib/client.ml
gsebil08: Agreed. I feel like it is more explicit to do this way. Thanks!

lib/client.ml
gsebil08: My bad, I forgot to revert it.
Also added the definition of `MSG_PEEK` flag.

lib/client.ml
rosalogia: If you want to change other truncated names to match your preference, I'm fine with that :)

gsebil08: I will do it later, this is not important right now.

rosalogia:
Looks good :) I'm going to squash and merge this PR, but in the future know that if your commit is simply "fixed the last commit", you can do `git commit --amend`. This rewrites history on the branch, so you will have to follow with `git push -f`. We don't want to use it on main because of this aspect. If you don't like this aspect (rewriting history), you can instead do something like `git commit --fixup HEAD` which will create a commit with the message "fixup! <message of last commit". This is fine as well.

rosalogia (approved):

PR #4, opened 2022-03-10
project(template): add templates for bug and feature issue creation

gsebil08:
First step of https://hackmd.io/9lg7TzVURoGWmlHlQ4bXuw => template for issue creation.


rosalogia (approved):
These look fine, we can adjust if after using them for a while we feel changes are necessary

PR #5, opened 2022-03-11
swim(initSwimProtocol): Failure detector module from SWIM protocol

gsebil08:
# Context

Most of the case, while using UDP we do not care about the reception (or not) of the sent messages. In case the acknowledge is important (to don't say mandatory) for your use case, you can develop your own acknowledge system and update the recipient regarding their response (or lack of response).

To efficiently address the Byzantine General Problem, we decided to implement the [SWIM protocol](https://www.cs.cornell.edu/projects/Quicksilver/public_pdfs/SWIM.pdf).

You can find a lot of readings about this protocol and more generaly about the gossip protocols, here are some we based our work on:
- [Correctness of a Gossip Based Membership Protocol](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.1088.9609&rep=rep1&type=pdf)
- [Diffusion without false rumors: on propagating
updates in a Byzantine environment](https://citeseerx.ist.psu.edu/viewdoc/download?doi=10.1.1.101.1404&rep=rep1&type=pdf)
- [On Byzantine Broadcast in Loosely Connected Networks](https://hal.sorbonne-universite.fr/hal-00728337/document)
- [A Survey and Comparison of
Peer-to-Peer Overlay Network Schemes](https://snap.stanford.edu/class/cs224w-readings/lua04p2p.pdf)
- [The Promise, and Limitations, of Gossip Protocols](https://research.cs.cornell.edu/projects/Quicksilver/public_pdfs/2007PromiseAndLimitations.pdf)

# What's inside?

This Pull Request implements the first step of the SWIM protocol: The failure detection.

This module contains all the functions to start the SWIM protocol on your side. Something like:
```ocaml
let protocol : Failure_detector.t =
  {
    config = { protocol_period = 5; round_trip_time = 2; peers_to_ping = 1 };
    acknowledges = Base.Hashtbl.Poly.create ();
    sequence_number = 0;
  }

Lwt.async (fun () -> failure_detection protocol client);
```
should do the job.


lib/peer.ml
```diff
@@ -24,3 +24,36 @@ let from_sockaddr sockaddr =
 
 let to_sockaddr peer =
   Lwt_unix.ADDR_INET (Unix.inet_addr_of_string peer.address, peer.port)
+
+(* Basic Knuth shuffle => https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle *)
+let knuth_shuffle known_peers =
+  let initial_array = Array.length (Array.of_list known_peers) in
+  let shuffled_array = Array.copy (Array.of_list known_peers) in
+  for i = initial_array - 1 downto 1 do
```
gsebil08, line 32: Is there any `functional` way to write it?

test/peer_tests.ml
```diff
@@ -0,0 +1,38 @@
+open Lwt.Infix
+open Commons
+
+module Peer_tests = struct
+  open Pollinate
+  let add_peer_test () =
+    let open Commons in
+    let peer_b = Peer.add_peer peer_a peer_b in
```
gsebil08, line 8: I think this is not correct. Actual `add_peer` function return a new peer, should it work this way or should I use `ref`?

lib/client.ml
```diff
@@ -7,6 +7,7 @@ type 'a t = {
   state : 'a ref;
   recv_mutex : Lwt_mutex.t;
   state_mutex : Lwt_mutex.t;
+  know_clients : 'a t list;
```
rosalogia, line 10: Why not Peer.t list?

lib/client.ml
```diff
@@ -31,7 +32,7 @@ let naive_broadcast client payload peers =
 let recv_next client =
   let open Lwt_unix in
   let open Util in
-  (* Peek at the first 8 bytes of the incoming datagram
+  (* Peek the first 8 bytes of the incoming datagram
```
rosalogia, line 35: Keep "Peek at the ..."

lib/peer.ml
rosalogia: Define a swap function on indices i and j, then Array.iteri to swap each index i with a random index j?

lib/peer.ml
```diff
@@ -24,3 +24,36 @@ let from_sockaddr sockaddr =
...
+    let x = shuffled_array.(k) in
+    shuffled_array.(k) <- shuffled_array.(i);
+    shuffled_array.(i) <- x
+  done;
+  Array.to_list shuffled_array
+
+(* Regarding the SWIM protocol, the list of peers is not ordered.
+   Hence, I basically went for a shuffle after adding the new peer *)
+let add_peer peer_to_add peer =
+  let new_peers = peer_to_add :: peer.known_peers in
```
rosalogia, line 43: Come to think of it, we should probably store peers in a data structure that allows us to look up a peer by address in estimated O(1) time. How about OCaml Map?

lib/peer.mli
```diff
@@ -1,9 +1,13 @@
...
+
+(** Sum type holding all the available values for peer state *)
 type state =
   | Alive
   | Suspicious
   | Faulty
 
-(** Our representation of a Peer *)
+(** Our representation of a Peer
+  See Client.peer_from to construct one *)
```
rosalogia, line 10: I like having notes like this, good idea

lib/client.ml
gsebil08: Oh, I don't know why I didn't think about it. Done!

lib/client.ml
gsebil08: Done.

lib/peer.mli
```diff
@@ -1,9 +1,13 @@
+(** Types and functions pertaining to peers in a P2P application *)
+
+(** Sum type holding all the available values for peer state *)
 type state =
```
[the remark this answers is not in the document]
gsebil08: You're right. Done.

lib/peer.ml
gsebil08: Totally reworked this function.

test/peer_tests.ml
gsebil08: Totally reworked it.

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+  round_trip_time : int;
+  timeout : int;
+}
+
+type t = {
+  config : config;
+  (* TODO: I am totally not sure about this type *)
+  acknowledges : (int, unit Lwt_condition.t) Poly.t;
+  mutable peers : Peer.t list;
+  (* TODO: I am not sure this is useful, it's just a counter to know what is the current round
```
rosalogia, line 20: The paper mentions using this, I think it's good and necessary

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+type config = {
+  protocol_period : int;
+  round_trip_time : int;
+  timeout : int;
+}
+
+type t = {
+  config : config;
+  (* TODO: I am totally not sure about this type *)
+  acknowledges : (int, unit Lwt_condition.t) Poly.t;
```
rosalogia, line 18: Can you explain what this is exactly?

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
+open Base.Hashtbl
+
+type message =
```
rosalogia, line 3: Given how Bin_prot encodes sum types, we should probably find a way to differentiate SWIM messages from the consumer's custom defined messages. One way to do this is to prepend some header bytes to the payload when packing SWIM messages and warn the consumer about this. What do you think?

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+  mutable sequence_number : int;
+}
+
+(* Helper to increase the round *)
+let next_seq_no t =
+  let sequence_number = t.sequence_number in
+  t.sequence_number <- t.sequence_number + 1;
+  sequence_number
+
+let send_ping_to client (peer_address : Peer.address) =
```
rosalogia, line 31: Probably can abstract into a helper function like

```ocaml
let send_msg msg client peer_address =
  let msg = Util.Encoding.pack bin_writer_message msg in
  Client.send_to client msg peer_address
```

So that we can do e.g.

```ocaml
let send_ping_to = send_msg Ping
```

rosalogia: This way we can also write any header-injection logic only once :+1:

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+  | PingRequest of Peer.address
+[@@deriving bin_io]
+
+type config = {
+  protocol_period : int;
+  round_trip_time : int;
+  timeout : int;
+}
+
+type t = {
```
rosalogia, line 15: `Swim.t` does not say a lot about what this object actually is. Maybe we can bypass the use of SWIM in the name and just call this module "Failure_detector" as in the paper. That way, `Failure_detector.t`'s purpose and functionality is more self-explanatory even if the consumer/reader of the code is unfamiliar with SWIM.

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+(* Regarding the SWIM protocol, the list of peers is not ordered.
+   Hence, I basically went for a shuffle after adding the new peer *)
+let add_peer peer_to_add t =
+  let new_peers = peer_to_add :: t.peers in
+  let shuffled_list = knuth_shuffle new_peers in
+  t.peers <- shuffled_list
+
+(* Regarding the SWIM protocol, if peer A cannot get ACK from peer B (timeout):
+   A sets B as `suspicious`
+   A randomly picks one (or several, should it also be randomly determined?) peer(s) from its list
```
rosalogia, line 78: Number of peers to choose each round can be configurable

rosalogia: One way to randomly choose x items from the list without chance of drawing the same item twice is to simply shuffle the list then take the first x items from the shuffled list.

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+  let _ =
+    update_peer t client
+      ({ status = Alive; socket_address = src_addr } : Peer.t) in
+  match msg with
+  | Ping -> send_acknowledge_to client src_addr
+  | PingRequest addr -> (
+    let seq_no = next_seq_no t in
+    let _ = send_ping_to client addr in
+    match%lwt wait_ack_timeout t seq_no t.config.protocol_period with
+    (* TODO: I am not sure about using the backtick character, I think it makes it polymorphic, should we do something like that here? *)
```
rosalogia, line 99: You can use `('a, 'b) Result.t` if you want. This is fine, but it can make type signatures a little weird.

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+   and ask him/them to ping B.*)
+(* This function return the random peer, to which we will ask to ping the first peer *)
+let pick_random_peer t =
+  let random_int = Random.int (List.length t.peers) - 1 in
+  List.nth t.peers random_int
+
+(* TODO: This function will update the status of the peer to Suspicious or Faulty *)
+let update_peer t client peer = failwith "undefined"
+
+(* How should I use this function? Let the client of the library defines a `peer.init` using it at `msg_handler`? *)
```
rosalogia, line 88: We will configure the default routing mechanism to scan the incoming payload and send it to this function if it's confirmed to be a SWIM-related message.

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+let handle_payload t client src_addr msg =
+  let _ =
+    update_peer t client
+      ({ status = Alive; socket_address = src_addr } : Peer.t) in
```
rosalogia, line 92: Can also do `Peer.{ status = Alive; socket_address = src_addr }`

lib/swim.ml
```diff
@@ -0,0 +1,127 @@
...
+  match msg with
+  | Ping -> send_acknowledge_to client src_addr
+  | PingRequest addr -> (
+    let seq_no = next_seq_no t in
```
rosalogia, line 96: I have to look at the paper again to remind myself when we should update sequence number. This comment is a reminder to myself to do so :smile:.

lib/peer.ml
```diff
@@ -1,26 +1,27 @@
-type state =
+open Bin_prot.Std
+
+type status =
   | Alive
   | Suspicious
   | Faulty
 
-type t = {
+type address = {
```
rosalogia, line 8: I agree with the creation of the address type so we can pack/unpack the address and send it around, but I think it makes for a cleaner API if this type isn't exposed and instead of other functions accepting `Peer.address`, they just accept `Peer.t` and use some function to access the address as necessary or construct the peer from an address.

lib/peer.ml
```diff
@@ -1,26 +1,27 @@
...
   address : string;
   port : int;
-  known_peers : t list;
```
rosalogia, line 9: Is there an issue with having the known peers list as part of Peer.t?

rosalogia (commented):
Looks like you've gotten a lot done. Good work :smile:. Once I finally finish debugging this issue and land the routing functionality, we can meet to talk about how this code will integrate with the server. Thanks again for your work :slightly_smiling_face:.

lib/swim.ml
gsebil08: Yup, nice idea! Done!

lib/swim.ml
gsebil08: Oh yes, I totally forgot that!
Switched to number of peers instead of only one. Thanks.

lib/swim.ml
gsebil08: Oh thanks! Much better, even if I still don't have the implementation of `wait_ack_timeout`.

lib/swim.ml
gsebil08: Thanks, switched to your proposal. Is it the "standard"?

rosalogia: I've seen it in Deku code 😅

lib/swim.ml
gsebil08: Totally agreed, renamed to `Failure_detector`!

lib/swim.ml
gsebil08: SO let's go for the same!

lib/swim.ml
gsebil08: In my mind, I just want to store:
(sequence_number, status) for each Peer

lib/client.ml
```diff
@@ -49,7 +47,7 @@ let recv_next client =
   (* Now that we have read the header and the message size, we can read the message *)
   let%lwt _, addr = recvfrom !client.socket msg_buffer 0 msg_size [] in
   Lwt_mutex.unlock !client.recv_mutex;
-  Lwt.return (msg_buffer, Peer.from_sockaddr addr)
+  Lwt.return (msg_buffer, peer_from (Address.from_sockaddr addr))
```
gsebil08, line 50: This is following your comment to switch back to `Peer.t` instead of `Address.t` everywhere.

lib/failure_detector.ml
```diff
@@ -0,0 +1,131 @@
...
+  protocol_period : int;
+  round_trip_time : int;
+  timeout : int;
+  mutable peers_to_ping : int;
+}
+
+type t = {
+  config : config;
+  (* TODO: I am totally not sure about this type *)
+  acknowledges : (int, unit Lwt_condition.t) Base.Hashtbl.t;
```
gsebil08, line 19: Just a comment to don't lose the conversation about this type holding something like (sequence, status) for each peer.

lib/client.ml
```diff
@@ -67,6 +65,14 @@ let init ~state ~msg_handler (address, port) =
   let state = ref state in
   let recv_mutex = Lwt_mutex.create () in
   let state_mutex = Lwt_mutex.create () in
-  let client = ref { address; port; socket; state; recv_mutex; state_mutex } in
+  let client =
+    ref
+      {
+        address = Address.create_address address port;
```
rosalogia, line 71: Since it's in Address module, we can just call this function `create`. Seems to be a common idiom.

lib/failure_detector.ml
rosalogia: Hmm, I think I understand better what this is now.

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+  | Ping
+  | Acknowledge
+  | PingRequest of Address.t
+[@@deriving bin_io]
+
+type config = {
+  protocol_period : int;
+  round_trip_time : int;
+  timeout : int;
+  mutable peers_to_ping : int;
```
rosalogia, line 13: This field doesn't have to be mutable, we can just make the config field in `t` mutable. Then, if we want to change the number, we can do:

```ocaml
let failure_detector : Failure_detector.t = ... in
failure_detector.config <- { failure_detector.config with peers_to_ping = 5 };
```

Unless, for some reason, we want only this config value to be mutable?

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+let send_acknowledge_to client recipient =
+  send_message client recipient Acknowledge
+
+let send_ping_request_to client recipient =
+  send_message client recipient (PingRequest recipient.socket_address)
+
+(* TODO: Totally not sure about this...
+   I don't know if I can get rid of it and only use `wait_ack_timeout` instead
+    But I feel like Lwt_condition allows ms to manage async thread and could be useful *)
+let[@warning "-32"] [@warning "-39"] rec wait_ack t sequence_number =
```
rosalogia, line 45: Why this function is recursive? Are you sure you don't want to do anything after `Lwt_condition.wait` returns?

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+
+(* TODO: An async thread should be started here, with a basic sleep, but not sure at all...
+     And we should return Either Ok Timeout *)
+let wait_ack_timeout t sequence_number timeout =
+  Lwt.pick
+    [
+      (let _ = Lwt_unix.sleep (Float.of_int timeout) in
+       Lwt.return @@ Result.Error "timeout");
+      (let _ = wait_ack t sequence_number in
+       Lwt.return @@ Result.ok "received ACK");
```
rosalogia, line 59: Above you use the `Error` constructor, here you use `ok` function. Stick to either constructors or functions here :+1:

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+  | elem :: list ->
+    if number_of_peers = 1 then
+      [elem]
+    else
+      elem :: pick_random_peers list (number_of_peers - 1)
+
+(* TODO: This function will update the status of the peer to Suspicious or Faulty *)
+let update_peer _t _client _peer = failwith "undefined"
+
+(* How should I use this function? Let the client of the library defines a `peer.init` using it at `msg_handler`? *)
```
rosalogia, line 98: Once this is merged and my PR is merged, I will open an issue and PR for integrating the failure detection protocol into the client, and I will handle that, so don't worry too much about this :smile:. If you would like, though, I will happily take time to show you how that will work. In fact, maybe best if we meet 1-on-1 next week to answer questions about each other's code and show examples etc.

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+    (* TODO: I am not sure about using the backtick character, I think it makes it polymorphic, should we do something like that here? *)
+    | Ok _ -> Lwt.return ()
+    | Error _ -> send_acknowledge_to client peer)
+  | Acknowledge ->
+  match find t.acknowledges t.sequence_number with
+  | Some cond ->
+    Lwt_condition.broadcast cond ();
+    Lwt.return ()
+  | None ->
+    (* Unexpected ACK -- ignore *)
```
rosalogia, line 118: If we get an ACK from a peer we didn't know before, can we actually just add it to the list of peers?

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+    Lwt.return ()
+
+(* TODO: This function should be call in a never endling loop to send the message each N seconds/minutes
```
rosalogia, line 121: Will be addressed during integration into client :+1:

lib/failure_detector.mli
```diff
@@ -0,0 +1,38 @@
+(** Messages sent by the protocol *)
```
rosalogia, line 1: Let's add a module description as well

lib/peer.ml
```diff
@@ -1,26 +1,12 @@
...
   | Alive
   | Suspicious
   | Faulty
 
 type t = {
-  address : string;
-  port : int;
-  known_peers : t list;
-  state : state;
+  socket_address : Address.t;
```
rosalogia, line 7: Maybe just address is fine for the field name

test/failure_detector_tests.ml
```diff
@@ -0,0 +1,44 @@
...
+  let add_peer_test () =
+    let _ = Failure_detector.add_peer Commons.peer_a Commons.protocol in
+    Lwt.return @@ List.length Commons.protocol.peers
+
+  let pick_random_test () =
+    let _ = Failure_detector.add_peer Commons.peer_a Commons.protocol in
+    let peers = Failure_detector.pick_random_peers Commons.protocol.peers 1 in
+    Lwt.return @@ List.length peers
+
+  let knuth_shuffle_test () =
```
rosalogia, line 15: Wonder if we can make these property-based tests

rosalogia (changes requested):
I think my review is complete :smile:. Looking forward to merging this; good work so far.

rosalogia (changes requested):
I think my review is complete :smile:. Looking forward to merging this; good work so far.

lib/client.ml
```diff
@@ -1,26 +1,24 @@
...
-let peer_from (client : 'a t) =
-  Peer.
-    {
-      address = client.address;
-      port = client.port;
-      known_peers = [];
-      state = Alive;
-    }
+let address_of client =
+  match client with
```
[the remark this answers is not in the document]
gsebil08: Done 👍

lib/client.ml
```diff
@@ -1,26 +1,24 @@
...
+  | t -> t.address
+
+let peer_from (address : Address.t) =
```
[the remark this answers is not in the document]
gsebil08: Oh yes, my bad, I forgot to cleanup my mess!
Totally agreed, done.

lib/client.ml
gsebil08: Nice catch! Done.

lib/failure_detector.ml
gsebil08: I got your point, but I am not fond of making the whole `type t` mutable when there is only one field which must be mutable.
What do you think?

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+  mutable sequence_number : int;
+}
+
+(* Helper to increase the round *)
+let next_seq_no t =
+  let sequence_number = t.sequence_number in
+  t.sequence_number <- t.sequence_number + 1;
+  sequence_number
+
+let send_message client (recipient : Peer.t) message =
```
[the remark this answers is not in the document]
gsebil08: Great! Eta reduce FTW!!

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
+open Base.Hashtbl
```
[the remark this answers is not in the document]
gsebil08: I went for explicit `Base.Hashtbl` as they are only two functions from this module.

lib/failure_detector.ml
gsebil08: Hum, I don't remember why I made it recursive at first... I just removed it.

lib/failure_detector.ml
gsebil08: Nice catch! Went fo constructor.

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+    But I feel like Lwt_condition allows ms to manage async thread and could be useful *)
+let[@warning "-32"] [@warning "-39"] rec wait_ack t sequence_number =
+  let cond =
+    find_or_add t.acknowledges sequence_number ~default:Lwt_condition.create
+  in
+  Lwt_condition.wait cond
+
+(* TODO: An async thread should be started here, with a basic sleep, but not sure at all...
+     And we should return Either Ok Timeout *)
+let wait_ack_timeout t sequence_number timeout =
```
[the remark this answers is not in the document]
gsebil08: Thanks! It takes me sooo long... But I think it is a good way.

lib/failure_detector.ml
gsebil08: Got it! Removed the comment.

lib/peer.ml
gsebil08: Agreed, done.

lib/failure_detector.mli
```diff
@@ -0,0 +1,38 @@
...
+
+(** Add the provided peer to the known peers from the protocol *)
+val add_peer : Peer.t -> t -> unit
+
+(** At each round, the protocol will randomly pick two peers:
+  - First one: will be the sender of the Ping
+  - Second one: the recipient *)
+val pick_random_peers : Peer.t list -> int -> Peer.t list
+
+(** Basic random shuffle, see https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle*)
```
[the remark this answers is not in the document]
gsebil08: For testing purposes I don't have the choice 😕

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+      (let _ = Lwt_unix.sleep (Float.of_int timeout) in
+       Lwt.return @@ Result.Error "timeout");
+      (let _ = wait_ack t sequence_number in
+       Lwt.return @@ Result.ok "received ACK");
+    ]
+
+(* Basic Knuth shuffle => https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle *)
+let knuth_shuffle known_peers =
+  let initial_array = Array.length (Array.of_list known_peers) in
+  let shuffled_array = Array.copy (Array.of_list known_peers) in
```
ulrikstrid, line 65: It's probably "expensive" to convert the same list to an array twice, you probably want to use `Array.copy initial_array` instead.

lib/failure_detector.ml
rosalogia: I don't think I understand why only this value is mutable 🤔

lib/failure_detector.ml
rosalogia: This is fine, we don't want to crash just because of that 😅

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+    [
+      (let _ = Lwt_unix.sleep (Float.of_int timeout) in
+       Lwt.return @@ Result.Error "timeout");
+      (let _ = wait_ack t sequence_number in
+       Lwt.return @@ Result.ok "received ACK");
+    ]
+
+(* Basic Knuth shuffle => https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle *)
+let knuth_shuffle known_peers =
+  let initial_array = Array.length (Array.of_list known_peers) in
```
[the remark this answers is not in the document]
rosalogia: Anyway, this will look a little different with hashtbl for peers so let's see then

lib/client.ml
gsebil08: Done!

lib/failure_detector.ml
```diff
@@ -0,0 +1,138 @@
...
+  let new_peers = peer_to_add :: t.peers in
+  let shuffled_list = knuth_shuffle new_peers in
+  t.peers <- shuffled_list
+
+(* Regarding the SWIM protocol, if peer A cannot get ACK from peer B (timeout):
+   A sets B as `suspicious`
+   A randomly picks one (or several, should it also be randomly determined?) peer(s) from its list
+   and ask him/them to ping B.*)
+(* This function return the random peer, to which we will ask to ping the first peer *)
+let rec pick_random_peers (peers : Peer.t list) number_of_peers =
```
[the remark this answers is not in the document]
gsebil08: Reworded it to shuffle the list of addresses and return this shuffled list, since we only need the address.

lib/failure_detector.ml
gsebil08: Nice catch! 👍 
changed to:
```ocaml
  let shuffled_array = Array.copy (Array.of_list known_peers) in
  let initial_array_length = Array.length shuffled_array in
```
to avoid duplicate `of_list`

lib/failure_detector.ml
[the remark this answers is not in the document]
gsebil08: Not changed at all, I am now shuffling the list of addresses instead of the list of peers.

lib/failure_detector.ml
gsebil08: Actually, after reading (again) the SWIM paper, I feel like this is not useful to make this field configurable from the client.
It looks like this number is "randomly decided by the protocol" and can be `internal`, what do you think?

lib/address.ml
```diff
@@ -0,0 +1,27 @@
+open Bin_prot.Std
+
+type t = {
+  address : string;
+  port : int;
+}
+[@@deriving bin_io, eq]
+
+let address_of address =
```
rosalogia: Hmm, since we are providing these functions and the address type is not mutable, what if we just expose the fields of the type in the mli? Seems like no harm would be done if we did

lib/address.ml
```diff
@@ -0,0 +1,27 @@
...
+  match address with
```
rosalogia: This match is unnecessary, can just do t.address or

```ocaml
let address_of { address ; _ } =
  address
```

lib/failure_detector.ml
rosalogia: Sounds good; some things are supposed to be configurable though, right? What were those?

lib/failure_detector.ml
```diff
@@ -0,0 +1,135 @@
...
+  t.sequence_number <- t.sequence_number + 1;
+  sequence_number
+
+let send_message message client (recipient : Peer.t) =
+  let message = Util.Encoding.pack bin_writer_message message in
+  Client.send_to client message recipient
+
+let send_ping_to client peer =
+  match peer with
+  | None -> failwith "Trying to send message to nobody)"
```
rosalogia, line 32: Why is peer optional?

lib/failure_detector.ml
```diff
@@ -0,0 +1,146 @@
...
+   A sets B as `suspicious`
+   A randomly picks one (or several, should it also be randomly determined?) peer(s) from its list
+   and ask him/them to ping B.*)
+(* This function return the random peer, to which we will ask to ping the first peer *)
+let rec pick_random_peer_addresses (peers : (Address.t, Peer.t) Base.Hashtbl.t)
+    number_of_peers =
+  let addresses = knuth_shuffle @@ Base.Hashtbl.keys peers in
+  match addresses with
+  | [] -> failwith "pick_random_peers"
+  | elem :: _ ->
```
rosalogia, line 87: If you'd like, the OCaml idiom here is "piping"

```ocaml
let addresses =
  peers
  |> Base.Hashtbl.keys
  |> knuth_shuffle
```

We tend to enjoy that this makes the flow of data through transformations very obvious 😄. OCaml format might reduce this to one line, but that's okay too.

lib/failure_detector.ml
```diff
@@ -0,0 +1,146 @@
...
+    if number_of_peers = 1 then
+      [elem]
+    else
+      elem :: pick_random_peer_addresses peers (number_of_peers - 1)
+
+(* Set the  *)
+let update_peer (peer : Peer.t) (peer_to_update : Peer.t) (status : Peer.status)
```
rosalogia, line 94: Cool 🙂

lib/failure_detector.ml
```diff
@@ -0,0 +1,146 @@
...
+    let _ =
+      send_ping_to client
+        (Peer.retrieve_peer_from_address_opt peer_from.peers addr) in
+    match%lwt wait_ack_timeout t new_seq_no t.config.protocol_period with
+    | Ok _ -> Lwt.return ()
+    | Error _ -> send_acknowledge_to client peer)
+  | Acknowledge ->
+  match Base.Hashtbl.find t.acknowledges t.sequence_number with
+  | Some cond ->
+    Lwt_condition.broadcast cond ();
```
rosalogia, line 121: What's warning 32?

lib/failure_detector.ml
```diff
@@ -0,0 +1,146 @@
...
+  | elem :: _ ->
+    if number_of_peers = 1 then
+      [elem]
+    else
+      elem :: pick_random_peer_addresses peers (number_of_peers - 1)
+
+(* Set the  *)
+let update_peer (peer : Peer.t) (peer_to_update : Peer.t) (status : Peer.status)
+    =
+  let updated_peer =
```
rosalogia, line 96: Incomplete comment

lib/failure_detector.mli
```diff
@@ -0,0 +1,37 @@
...
+  (* NUmber of peers to pick at each round *)
+  peers_to_ping : int;
+}
+
+(** Type holding all the necessary information for the SWIM protocol *)
+type t = {
+  config : config;
+  acknowledges : (int, unit Lwt_condition.t) Base.Hashtbl.t;
+  mutable sequence_number : int;
+}
```
rosalogia, line 22: Hashtbl is a mutable structure itself, so it doesn't need to be marked mutable 👍 only needs to be mutable if you need to use <- with it

lib/peer.ml
```diff
@@ -1,26 +1,24 @@
...
 
 type t = {
-  address : string;
-  port : int;
-  known_peers : t list;
-  state : state;
+  address : Address.t;
+  mutable status : status;
+  mutable peers : (Address.t, t) Base.Hashtbl.t;
 }
```
rosalogia, line 11: I think this belongs in the failure detector internally. The peer type isn't really concerned with this type of hashtbl.

lib/peer.ml
```diff
@@ -1,26 +1,24 @@
...
-  | ADDR_INET (inet, port) ->
-    {
-      address = Unix.string_of_inet_addr inet;
-      port;
-      known_peers = [];
-      state = Alive;
-    }
+let retrieve_peer_from_address_opt (peers : (Address.t, t) Base.Hashtbl.t)
+    address =
+  Base.Hashtbl.find peers address
```
rosalogia, line 15: Why not call it `of_address` so it reads `Peer.of_address client.address` for example

lib/peer.ml
```diff
@@ -1,26 +1,24 @@
...
 
-let to_sockaddr peer =
-  Lwt_unix.ADDR_INET (Unix.inet_addr_of_string peer.address, peer.port)
+let peer_from (address : Address.t) =
+  {
```
rosalogia, line 17: Can just remove peer_ at the beginning since the module name conveys this

lib/peer.mli
```diff
@@ -1,18 +1,24 @@
...
+
+(** Sum type holding all the available values for peer state *)
+type status =
   | Alive
   | Suspicious
   | Faulty
 
-(** Our representation of a Peer *)
+(** Our representation of a Peer
+  See Client.peer_from to construct one *)
```
rosalogia, line 10: Outdated comment right?

lib/failure_detector.ml
gsebil08: Every field must be configurable, but I think it is better to "force" the client to create a new `failure_detector.t`.
Maybe we can expose a `create` function to ease this.
Do you think it is a good idea?

lib/address.ml
gsebil08: Yup, done this morning 😉

lib/failure_detector.ml
gsebil08: Thanks! Don't have (yet) the habit to use `|>`!

lib/failure_detector.ml
gsebil08: Unused function

lib/failure_detector.ml
gsebil08: 😇 done

lib/failure_detector.mli
gsebil08: Oh, seems obvious since I used `update` function... Thanks

lib/peer.ml
gsebil08: Thanks, seems obvious once again.

lib/peer.mli
gsebil08: Removed.

lib/peer.ml
rosalogia: Referring here to `retrieve_peer_from_address_opt`

gsebil08: Actually, it is. Since a peer must knows a list of peers. I went for `Hashtbl` for convenience.

test/failure_detector_tests.ml
```diff
@@ -0,0 +1,44 @@
...
+open Lwt.Infix
+
+module Failure_detector_tests = struct
+  open Pollinate
+  let add_peer_test () =
+    let _ = Failure_detector.add_peer Commons.peer_a Commons.protocol in
+    Lwt.return @@ List.length Commons.protocol.peers
+
+  let pick_random_test () =
+    let _ = Failure_detector.add_peer Commons.peer_a Commons.protocol in
```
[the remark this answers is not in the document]
gsebil08: Added.

lib/client.ml
```diff
@@ -1,26 +1,27 @@
...
-  Peer.
-    {
-      address = client.address;
-      port = client.port;
-      known_peers = [];
-      state = Alive;
-    }
+let address_of { address; _ } = address
+
+let peer_from { address; _ } =
```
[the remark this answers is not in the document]
rosalogia: `to_peer` might be better, since we are turning a client into a peer. `Client.to_peer client_a` returns a peer with the address of `client_a`, I think that makes sense.

lib/peer.ml
rosalogia: Yeah, my comment got misplaced :sweat_smile: you're right, what I suggest doesn't make sense for where this comment ended up.

rosalogia:
@gsebil08 Good work on this PR :smile: We are certainly and undoubtedly almost done. There are some minor changes I would like to see, and to expedite merging of this PR, I'm going to try making them and explaining each one. Feel free to rebase and omit changes you don't agree with.

rosalogia:
I'm going to move some functions from Failure_detector which act solely on Peers and their data to the Peer module, this should clean up the Failure_detector module a little and allow us to document all Peer functionalities in one place. Let me know if you're not in favor of this change :+1:.

rosalogia:
Question: is sequence_number actually the same as sequence number in the paper? Or does it correspond better to the notion of an "incarnation number"? "Sequence number" is mentioned once in the paper, and its purpose is not at all specified. The purpose of "incarnation numbers" is much more clear, and it seems like that's what we want to achieve with sequence numbers in our code. However, it is important, then, that the incarnation number be stored alongside each peer.

rosalogia:
For now, since most of the functions tested are internal utilities that should not be exposed, I'm going to remove them from the mli and comment out the tests. In the future, we can test the primary functionality of the failure detector, and maybe we can use in-line tests just to make sure there is no fault in the intermediate functions if necessary.

rosalogia:
Okay, I made the changes I saw as worth making before merging. I'm ready to merge now, but you may check out the changes and let me know if you disagree with any of the changes. Good night :smile:.

rosalogia (approved):

lib/failure_detector.ml
```diff
@@ -17,6 +17,9 @@ type t = {
   mutable sequence_number : int;
 }
 
+let create config =
```
gsebil08, line 22: Totally agreed with that!
Thanks.

lib/address.ml
```diff
@@ -0,0 +1,23 @@
...
+open Hash.Builtin
+open Ppx_sexp_conv_lib.Conv
+
+type t = {
+  address : string;
+  port : int;
+}
+[@@deriving bin_io, eq, compare, hash, sexp, show { with_path = false }]
+
+let create address port = { address; port }
```
ulrikstrid, line 13: The conventional babe for these functions in OCaml is `make`

lib/client.mli
```diff
@@ -8,7 +8,10 @@ receive a message (sending is non-blocking), or when the
 state is being read from or written to by the server. *)
 type 'a t
 
-(** Constructs a Peer.t from an 'a Client.t*)
+(** Retrieve the Address.t of the given Client.t *)
+val address_of : 'a t -> Address.t
```
ulrikstrid, line 12: Conventional name on OCaml would be `to_address`

ulrikstrid (commented):
Added some notes about conventional names, feel free to ignore

gsebil08:
@rosalogia 
> I'm going to move some functions from Failure_detector which act solely on Peers and their data to the Peer module

⬆️  Totally agreed, thanks!

>For now, since most of the functions tested are internal utilities that should not be exposed, I'm going to remove them from the mli and comment out the tests.

⬆️ Ok, I open an issue to track this and rework correctly the tests suite.

## Sequence number

>is sequence_number actually the same as sequence number in the paper? Or does it correspond better to the notion of an "incarnation number"?

It is the `sequence_number` mentioned in the paper:
>The data contained in each message of this protocol is tagged with the unique sequence number of the protocol period at the initiator

I just didn't add it _(yet)_ to each sent message, since I think we need your actual PR #8 before going further.

## Incarnation number

From what I understood, the `incarnation number` is initialized to 0 when a peer `join` the protocol. And each time the status peer is changed, its incarnation number must be incremented (section 4.2 of [this document](https://www.cs.cornell.edu/projects/Quicksilver/public_pdfs/SWIM.pdf))

> From the above discussion, Alive messages override Suspect messages, and Confirm messages override both Suspect and Alive messages, in their effect on the local membership list element corresponding to the suspected member.
>However, a member might be suspected and unsuspected multiple times during its lifetime.
> These multiple versions of Suspect and Alive messages (all pertaining to the same member) need to be distinguished through unique identifiers.
> These identifiers are provided by using a virtual incarnation number field with each element in the membership lists. Incarnation numbers are global.
> A member ’s incarnation number is initialized to 0 when it joins the group, and it can be incremented only by, when it receives information (through the Dissemination Component) about itself being suspected in the current incarnation - then generates an Alive message with its identifier and an incremented incarnation number, and spreads this through the Dissemination Component to the group.

lib/address.ml
gsebil08: Thanks, done.

lib/client.mli
gsebil08: Kinda disagree. I think it was ok with the previous comment (which was obviously false), but here, we `extract` the `Address.t` of the provided `Client`.
`to_address` implies we are either transforming a `Client.t` to an `Address.t`, or creating a new `Address.t` from the provided `Client.t`, which is not the case here. This function is only a `getter`.

gsebil08:
@rosalogia from my point of view, this is a solid base to rely on.
Of course this is not perfect:
- sequence number is not sent within each message
- incarnation number is not present at all
- `Suspicion Mechanism` is not implemented (see the `Faulty` hard coded within `failure_detection`)
- Almost no tests on exposed functions

But it is already a big PR, and I don't want it to become a massive one, which would be really hard to review.

lib/address.ml
rosalogia: IIRC, `make` is for functors (e.g. Map.Make, Set.Make) and `create` is more common for e.g. `Queue.create`, `Bytes.create`, etc. Not going to block on this though 👍

rosalogia:
LGTM‼️

PR #7, opened 2022-03-14
Init nix setup

ulrikstrid:
Bootstrapped with `nix flake init --template github:ulrikstrid/nix-templates#ocaml`

### DIFF
--- a/.envrc
+++ b/.envrc
@@ -1,0 +1,1 @@
+use flake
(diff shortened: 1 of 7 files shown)


ulrikstrid:
@gsebil08 if you want to use nix I have now rebased this PR so please review. If you don't want it please close it :smile:

gsebil08 (changes requested):
Would you mind play `nixfmt` on nix folder?
Also, please add `.direnv` folder to .gitignore.

Otherwise, this works like a charm!

gsebil08 (approved):

PR #8, opened 2022-03-16
Fix server intercepting responses to client requests

### LINKED ISSUE #2: Server sometimes receives and misinterprets request response
Currently the server and the client share a socket, and the server calls `recvfrom` asynchronously while the client is at liberty to make requests and receive responses to requests via a send and receive API. There are two big issues with this

1. When the client makes a request a peer, there's nothing stopping the server's call to `recvfrom` from receiving the response from the peer.
2. When serializing a value of a sum type (like our request and response types in our tests), Bin_prot does **not** preserve type information. Thus, in our tests, when the server unpacks the serialized payload, it has no way of knowing whether it's unpacking a response or a request.

Some ways we could resolve this include

1. Having a separate socket for the client and the server, I like this solution the least. I tried implementing it and it results in an annoying amount of code duplication.
2. Simply ensuring that requests and responses are not represented the way they currently are in our tests. One alternative that would work is simply wrapping requests and responses in a message sum type:
   ```ocaml
    type message =
      | Request of Request.t
      | Response of Response.t
    ```
   This way, whether or not the message is a request or a response would be encoded in the serialized payload. I don't feel too strongly about this solution either, since it places a restriction on the end user.
3. Creating a message type as specified above _in_ the library, but with requests and responses merely being `bytes`. This way we can ensure that for our purposes, whether the message is a request or a response is encoded in the serialized payload, but abstract this away from the consumer. This solution might be clean, but I'll have to give it a try and see.
4. If we really must, we can dive into the depths of the `Lwt_unix` module and see how we can set up polling and blocking, that might solve our problems but I honestly don't know yet :smile:

rosalogia:
## Problem

The server almost always receives any incoming message (via `recvfrom`) regardless of whether the message is intended to be processed by the server, or whether the message was meant as a response to a lone-request made by the client. Furthermore, due to how we were modelling requests and responses, and how Bin_prot serializes sum types, the server was interpreting the client-bound responses as server-bound requests without raising any kind of type-related error. This issue illuminated the greater issue of there being no mechanism in place for determining who gets to process an incoming message.

## Solution

Fixes #2 with the following solution:

* The server still receives any incoming message first, but a routing mechanism is in place such that a server only processes a message if the router determines that the message is a `Request`.
* A thread-safe queue library implemented using `Lwt_mutex` and `Lwt_condition` allows clients to store messages in "inboxes", one for requests and one for responses.
* When a message is received by the server, the message is processed by the router to determine which inbox the message should be sent to. This system can be extended to account for, e.g., the membership protocol that @gsebil08 is currently implementing.
* A `Client.request` function has been implemented, which takes advantage of `Tqueue.wait_to_take`, a function that blocks the current thread of execution until an element from the thread-safe queue is available to take. The `Client.request` function thus sends a request to a peer and blocks until a response appears in its response inbox.

Now, typical communication between nodes undergoes the following cycle:

1. A node, A, uses `Client.request` to send a request to a peer, B
2. Node B's server receives the message, and passes it to the consumer-provided _router_, which inspects the message to identify whether it is a request or a response, and possibly transforms it for later processing (e.g. removing unnecessary headers)
3. Node B's router determines that the message is a request, and thus the message is stored in Node B's `request_inbox`
4. Node B's consumer-provided _message handler_ takes the latest request from the inbox (which contains info. about the sender), and the current state of Node B as input and decides how to respond to Node A.
5. Node B's message handler packs its response to Node A with a binary encoding protocol, in our case Bin_prot.
6. Node B's server sends the response to Node A.
7. Node A's server receives the response from Node B and passes it to Node A's router, which determines that the message is a response. Thus, the message is stored in Node A's `response_inbox`.
8. Node A's call to `Client.request` resolves, since `Tqueue.wait_to_take` stops blocking execution now that there is an element in Node A's `response_inbox`. Node A may now use Node B's response however it wishes.

In code, the request is as simple as

```ocaml
let request : bytes = ... in
let%lwt response : Message.t = Client.request client_a request peer_b in
...
```


lib/client.ml
```diff
@@ -1,10 +1,24 @@
 open Lwt_unix
 
+module Message = struct
```
gsebil08, line 3: Should it be extracted to a dedicated module?

lib/client.ml
```diff
@@ -1,10 +1,24 @@
...
+  type label =
```
gsebil08, line 4: I think this type will be enrich with the ones I had defined on my PR, right?

lib/client.ml
```diff
@@ -1,10 +1,24 @@
...
+    payload : bytes;
+  }
+end
+
 type 'a t = {
   address : string;
   port : int;
   socket : file_descr;
   state : 'a ref;
+  request_inbox : Message.t Tqueue.t;
```
gsebil08, line 20: What about `inbox` and `outbox`?

lib/client.ml
```diff
@@ -51,22 +65,62 @@ let recv_next client =
...
-    Lwt_mutex.unlock !client.state_mutex;
+    let%lwt message, peer = recv_next client in
+    let%lwt () = route client peer message router in
+
+    let%lwt request = next_request client in
+
+    let%lwt () =
+      match request with
+      | Some request ->
+        let%lwt () = Lwt_mutex.lock !client.state_mutex in
```
gsebil08, line 92: What is the behavior of the lock? Will it lock only the current thread?

lib/tqueue.ml
```diff
@@ -0,0 +1,65 @@
+type 'a t = {
```
gsebil08, line 1: Why `Tqueue` instead of `Queue`? 🤔

lib/tqueue.ml
```diff
@@ -0,0 +1,65 @@
...
+      Queue.add x queue;
+      let%lwt () =
+        if Queue.length queue = 1 then (
+          Lwt_condition.signal has_elt ();
+          Lwt.return ())
+        else
+          Lwt.return () in
+      Lwt.return ())
+
+let push = add
```
gsebil08, line 24: What's the purpose of these? Why simply rename `add` to `push`? 🤔

lib/tqueue.ml
```diff
@@ -0,0 +1,65 @@
...
+      let%lwt () =
+        if Queue.is_empty queue then
+          Lwt_condition.wait ~mutex:lock has_elt
+        else
+          Lwt.return () in
+      Lwt.return (Queue.take queue))
+
+let pop = take
+
+let peek { queue; lock; _ } =
```
gsebil08, line 43: I think it is common to provide these two functions, but what about removing this one and only use the safe one?

gsebil08 (commented):
🎉 !!!

lib/client.ml
rosalogia: Yes

lib/client.ml
rosalogia: Yes!

lib/client.ml
rosalogia: `outbox` typically refers to a container of messages that are waiting to be sent :smile: neither of these really match that definition. However, I am not necessarily pleased with the current design of having one inbox field for every type of message.

lib/client.ml
rosalogia: The lock stops any other thread from reading from/writing to the client's state until the lock is released

lib/tqueue.ml
rosalogia: Queue's are not thread safe, I implemented a thread-safe queue library that uses mutexes to protect read/write operations, and a condition var to allow blocking until the queue has an element.

lib/tqueue.ml
rosalogia: Just copying https://ocaml.org/api/Queue.html for minimal friction :smile:

lib/tqueue.ml
rosalogia: Similar to above comment, I'm basically just copying [OCaml's Queue module](https://ocaml.org/api/Queue.html) but making it thread-safe and adding the functionality with condition vars to wait for an element to appear

lib/client.ml
gsebil08: You are right, `outbox` does not fit at all. Do you see any other kind of messages? I mean a peer can either send `request` or receive `response` so it looks like we will be stuck with these two for the rest of the life, am I right?

lib/client.ml
gsebil08: Is it like an `mvar` which is under the hood of the [Concurrent.Lock](https://hackage.haskell.org/package/concurrent-extra-0.7.0.12/docs/Control-Concurrent-Lock.html) I already used in Haskell? It would help me if it is the case 😇

lib/tqueue.ml
gsebil08: So maybe we can get rid of them, don't you think?

lib/tqueue.ml
gsebil08: Ok got it, I think we should get rid of the unsafe functions.

lib/tqueue.ml
rosalogia: Sure :slightly_smiling_face: I wasn't planning on using them

lib/message.mli
```diff
@@ -0,0 +1,19 @@
+(** Messages received by the client, whether they're requests,
+responses, or protocol-specific messages. For consumer use
+only when implementing a routing function for the
+client. *)
+
+(** Messages are requests or responses,
+determining how they're stored and
+where they're handled *)
+type label =
```
gsebil08, line 9: I am not totally convinced by this naming. I feel `message_type` would be better, what do you think?

lib/util.ml
```diff
@@ -30,6 +30,14 @@ module Encoding = struct
   let unpack reader payload =
     let buf = prepare_buff payload in
     read_all reader buf
+
+  let str_dump payload =
```
gsebil08, line 34: I know this function is a basic transformation, is it possible to add a comment after some important lines to explain what is the related step?
Or maybe just an example of `payload -> result`

test/client_tests.ml
```diff
@@ -1,62 +1,53 @@
...
-
-    let res_a, res_b =
-      match (res_a, res_b) with
+    let res_from_b, res_from_a =
+      match (res_from_b, res_from_a) with
       | List l1, List l2 -> (List.hd l1, List.hd l2)
       | _ -> failwith "Incorrect response" in
 
-    Lwt.return (res_a, res_b)
+    Lwt.return (res_from_b, res_from_a)
```
gsebil08, line 24: Smart naming change!

test/client_tests.ml
```diff
@@ -87,7 +76,7 @@ let () =
          ( "communication",
            [
              Alcotest_lwt.test_case "Trading Messages" `Quick test_trade_messages;
-             Alcotest_lwt.test_case "Ping pong" `Quick test_ping_pong
-             (* Alcotest_lwt.test_case "Insert value" `Quick test_insert_value; *);
+             Alcotest_lwt.test_case "Ping pong" `Quick test_ping_pong;
+             Alcotest_lwt.test_case "Insert value" `Quick test_insert_value;
```
gsebil08, line 80: 🎉 👏

test/commons.ml
```diff
@@ -16,37 +17,65 @@ module Commons = struct
     | Error   of string
   [@@deriving bin_io, show { with_path = false }]
 
+  type message =
```
gsebil08, line 20: Can't we reuse `message.ml` 🤔 ?

lib/util.ml
rosalogia: Can you explain a little more what you want me to comment on?

test/commons.ml
rosalogia: Consumer should have the freedom to define their own type for messages sent back and forth between nodes, this is an example of that. Message.t has peer information that the consumer shouldn't necessarily have to include in their messages.

lib/message.mli
rosalogia: How about category? In the rest of the codebase the type will appear `Message.message_type` which I'd like to avoid 😅

gsebil08: Ok for `category`.

lib/util.ml
gsebil08: Re-thinking about it, a comment is useless. At least a unitary test would be more helpful.

test/commons.ml
gsebil08: Yeah, sorry, I was mixed `payload` and `label`. My bad.

rosalogia:
Finally, we have something working that is, to my knowledge, not buggy :smile:. We have a mechanism for deciding where messages go (that is, who is responsible for handling them), and we have a mechanism for ensuring that `request` only returns the response to that request. We have convenience functions for creating request and response messages. Message.t is the new way to send or receive any kind of message, it contains some useful metadata for the library. Consumers don't actually have to worry about this because their interface is still just `bytes`. If they want to create a custom router (though the default is quite simple, sane and unsurprising), they have to work with `Message.t`, but that's fine.

This PR is way too big, so sorry about that :sweat_smile:. Right now, integration with failure detector is not there, but without the failure detector all the tests pass. Tomorrow I will integrate the failure detector on a branch off of this branch. In the meantime, although it might be a pain, please review this branch for me so that we can get it merged finally :slightly_smiling_face:.

rosalogia:
Tips for reviewing:

* Read the .mli file first
* Check out the branch on your local machine as well as looking at the diff in the browser, sometimes the diff can make it harder to read

lib/client.mli
```diff
@@ -1,38 +1,51 @@
...
+create functions to then feed to this function. *)
+val send_to : 'a t ref -> Message.t -> unit Lwt.t
 
-(** Initializes the client with an initial state and a message
-handler that acts on the current state, the peer sending the message,
-and the message itself in bytes. The message handler is used
+(** Waits for the next incoming message and returns it *)
+val recv_next : 'a t ref -> Message.t Lwt.t
+
+(** Sends an encoded request to the specified peer and
```
gsebil08, line 28: This is great!

lib/client.ml
```diff
@@ -1,75 +1,148 @@
...
+          request |> msg_handler state |> create_response client request in
+        let%lwt () = send_to client response in
+        Lwt.return (Mutex.unlock !client.state)
+      | None -> Lwt.return () in
+
     server () in
   Lwt.async server
 
-let init ~state ~msg_handler (address, port) =
+let init ~state ?(router = fun m -> m) ~msg_handler (address, port) =
```
gsebil08, line 135: I am not sure about this `fun m -> m` does it mean 
> I need a router which is a function taking a `m` and returning the same `m`?

If yes, in Haskell it would have been `const m` is there the same in OCaml?
If no, would you mind explain me?

lib/client.mli
```diff
@@ -1,38 +1,51 @@
 (** Types and functions pertaining to clients in a P2P application *)
 
-(** Clients are defined by a UDP socket bound
-to an address and port, a reference to some state
-whose type is user-defined, a mutex that gets
-locked whenever the client's socket is being used to
-receive a message (sending is non-blocking), or when the
-state is being read from or written to by the server. *)
+(** Represents a node with some state in a peer-to-peer network *)
```
gsebil08, line 3: Hum, we're introducing a new word: node.
Should it be replaced by `client` to avoid any confusion? Or is it a different thing?

lib/inbox.ml
```diff
@@ -0,0 +1,27 @@
...
+  |> find_or_create_category inbox
+  |> if consume then Tqueue.take else Tqueue.peek
+
+let await_next inbox ?(consume = true) category =
+  category
+  |> find_or_create_category inbox
+  |> if consume then Tqueue.wait_to_take else Tqueue.wait_to_peek
+
+let push inbox category message =
+  category |> find_or_create_category inbox |> Tqueue.add message
```
gsebil08, line 27: I do prefer:
```ocaml
let push inbox category message =
   Tqueue.add message @@ find_or_create_category inbox category
```
because I feel weird about the `category` in first.
Of course, if your syntax is more common, let's use it 😄

lib/inbox.mli
```diff
@@ -0,0 +1,25 @@
+(** A Hashtbl mapping message categories to thread-safe queues
+containing messages *)
+type t
+
+(** Initializes an inbox with queues for each type of message. *)
```
gsebil08, line 5:
```suggestion
(** Creates an inbox with queues for each type of message. *)
```

lib/message.ml
```diff
@@ -0,0 +1,17 @@
+open Bin_prot.Std
+
+type category =
```
gsebil08, line 3: I think we should only go for 
```ocaml
type category =
  | Request
  | Response
  | Failure_detector (* not sure about this one *)
  | Unsupported
[@@deriving bin_io]
```
what do you think?

gsebil08 (changes requested):
What a massive work! 💪 🧑‍💻 
I took hours to review, looks really great, thanks!
Would you mind play a 
```shell script
$ esy b dune build @fmt --auto-promote
```
before pushing?

lib/inbox.ml
```diff
@@ -0,0 +1,27 @@
+type t = (Message.category, Message.t Tqueue.t) Base.Hashtbl.t
+
+let create () =
+  let inbox = Base.Hashtbl.Poly.create () in
```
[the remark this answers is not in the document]
rosalogia: Otherwise I need a module for the key type

lib/message.mli
```diff
@@ -0,0 +1,25 @@
+(** Messages received by the client, whether they're requests,
+responses, or protocol-specific messages. For consumer use
+only when implementing a routing function for the
+client. *)
+
+(** Messages are requests or responses,
+determining how they're stored and
+where they're handled *)
+type category =
+  | Uncategorized
```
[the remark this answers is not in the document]
rosalogia: You're right, Custom is probably enough

lib/client.ml
```diff
@@ -1,75 +1,148 @@
...
+  Lwt.return message
+
+let request client request recipient =
+  let%lwt message = create_request client recipient request in
+  let%lwt () = send_to client message in
+  let condition_var = Lwt_condition.create () in
+  Hashtbl.add !client.request_table message.id condition_var;
+  Lwt_condition.wait condition_var
+
+let broadcast_request client req recipients =
```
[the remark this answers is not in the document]
rosalogia: I agree that `client.peers` should be a thing, but maybe it is good to have some fine-grain control over who you're sending a broadcast to.

lib/client.ml
rosalogia: It's the identity function yes, same as `const m`. We could use `Fun.id`but I'm not sure that's any more self explanatory. The purpose of `fun m -> m` here is that the router takes an incoming message as input, may change/decide its label, and may modify the payload before it's put into an inbox.

lib/client.mli
rosalogia: Honestly, maybe `node` is the better word than client? Makes the client/server terminology less confusing by not overloading client? Up to you.

lib/client.mli
```diff
@@ -1,38 +1,51 @@
...
+that responds to `request` whose content is `payload`. *)
+val create_response : 'a t ref -> Message.t -> bytes -> Message.t
 
-(** Waits for the next incoming datagram and returns the
-serialized payload along with the peer who sent the datagram *)
-val recv_next : 'a t ref -> (bytes * Peer.t) Lwt.t
+(** Sends a message via datagram from the given client
+to a specified peer. Construct a message with one of the
+create functions to then feed to this function. *)
+val send_to : 'a t ref -> Message.t -> unit Lwt.t
```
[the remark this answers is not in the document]
rosalogia: It depends on whether we want to allow consumers to define custom message types. If they do, then these functions will not suffice. Also you are right, the failure detector component needs access to `send_to`.

lib/inbox.ml
```diff
@@ -0,0 +1,27 @@
...
+
+let find_or_create_category inbox category =
+  Base.Hashtbl.find_or_add inbox category ~default:(fun () -> Tqueue.create ())
+
+let next inbox ?(consume = true) category =
+  category
+  |> find_or_create_category inbox
+  |> if consume then Tqueue.take else Tqueue.peek
+
+let await_next inbox ?(consume = true) category =
```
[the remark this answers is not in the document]
rosalogia: It was useful before, now I don't need it. Will remove and add again if it becomes necessary.

lib/inbox.mli
rosalogia: :+1:

lib/mutex.mli
```diff
@@ -0,0 +1,28 @@
...
+a mutex. *)
+type 'a t
+
+(** Creates a mutex-wrapped value *)
+val create : 'a -> 'a t
+
+(** Locks the mutex protecting the value and
+returns a reference to the value so that it can
+be mutated *)
+val lock : 'a t -> 'a Lwt.t
```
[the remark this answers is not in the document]
rosalogia: I think having a short name and good documentation + type signature explaining how it's used is okay in this case.

lib/tqueue.ml
rosalogia: They correspond to the safe versions of the functions now, so I don't see any issue with them.

lib/message.ml
gsebil08: Deku will be our first/main client. Hence, its scope/needs must be the only one we target for the beginning.
>[...] They will be forced to give up on using our library

I don't agree with that, they can still open an issue and ask us to handle a new `category`, don't you think?

lib/client.mli
gsebil08: Ok for `node`!

gsebil08 (approved):

lib/util.ml
gsebil08: Will be done within #19

gsebil08 (approved):

gsebil08:
Build is KO, but will be fixed with #21 as we wanted to split this work in several PR

Issue #10, opened 2022-03-16
[MAINTENANCE] Add proper logging support


rosalogia:
### Summary

We should use a strong logging library that provides us with the ability to hide log messages with log levels at the very least. Maybe `logs`?

### Relevant Documentation and Links

* [`logs` opam page](https://opam.ocaml.org/packages/logs/)
* [`logs` documentation](https://erratique.ch/software/logs/doc/)

oteku:
If [pollinate](https://github.com/marigold-dev/pollinate) remain a library, logging system must be a dependency injected at by the project which use pollinate.
Meaning it would be the purpose of deku to provide you a logging system that fit your expected interface (which is easy for loging)

This is usually achieved in OCaml using functors or virtual library

PR #19, opened 2022-03-28
WIP test(initPBT): switch some tests to PBT

### LINKED ISSUE #15: [MAINTENANCE]
### Summary

Some tests seems to be property based:
```ocaml
let test_knuth_shuffle _ () =
  Failure_detector_tests.knuth_shuffle_test ()
  >|= Alcotest.(check int) "Knuth_shuffle does not change length of list" 1
```

or even

```ocaml
let test_add_peer _ () =
  Failure_detector_tests.add_peer_test ()
  >|= Alcotest.(check int)
        "When adding peer to a empty list of know_peers, length is equal to 1" 1
```

It would be great to use `QCheck` to migrate them to PBT.

### Relevant Documentation and Links

- [QCheck documentation](https://blag.cedeela.fr/quickcheck-for-ocaml/)
- [PBT project example](https://github.com/miage-lille/troll-of-fame)

gsebil08:
Closes #15


test/generator.ml
```diff
@@ -0,0 +1,38 @@
...
+    {address = address; status = Alive; neighbors = Base.Hashtbl.Poly.create ()}
+
+  let peer_print peer =
+    let open Peer in
+    Address.show peer.address
+  
+  let insert_gen =
+    string_printable >|= fun str -> Commons.Insert str
+
+  let request_gen =
```
gsebil08, line 36: This generator is not working (see CI build fail)... I don't understand how I can make it works. I tried to use `QCheck2.Gen.join`, was compiling but tests was failing :/

test/util_prop.ml
```diff
@@ -0,0 +1,20 @@
...
+open Commons
+
+let pack_unpack =
+  QCheck2.Test.make 
+  ~count:1000
+  ~name:"unpack . pack returns the original value"
+  request_gen
+  (fun random_request -> 
+    let open Commons in 
+    random_request ==
```
rosalogia, line 12: In OCaml, we typically want `=` for equality checking, not `==`. https://stackoverflow.com/questions/13590307/whats-the-difference-between-equal-and-identical-in-ocaml

test/address_prop.ml
```diff
@@ -0,0 +1,37 @@
+open Pollinate
+open QCheck2.Gen
+open Generator.Test
+
+let create =
+  QCheck2.Test.make ~count:1000
+    ~name:"create from string and port is successful" (pair string int)
```
gsebil08, line 7: Which makes me think: should we crash of the address is not well formatted? (for example "address" should fail and "a.b.c.d.e" should also fail).
What do you think? Is it overkill?

test/client_tests.ml
```diff
@@ -1,14 +1,49 @@
 open Lwt.Infix
 open Pollinate
-open Pollinate.Util
-open Commons
+open Messages
 
 module Client_tests = struct
+  type state = string list
+
+  let protocol : Client.failure_detector_config =
```
gsebil08, line 8: This is not working

lib/node/node.ml
```diff
@@ -3,6 +3,7 @@ open Types
 
 module Message = Message
 module Client = Client
+module Failure_detector = Failure_detector
```
gsebil08, line 6: This is not working as expected. I thought I would be able to do `open Pollinate.Node.Failure_detector` in the tests suite, but nope...

lib/common/util.ml
```diff
@@ -45,7 +45,7 @@ end
 module Net = struct
   let create_socket port =
     let open Lwt_unix in
-    let ssock = socket PF_INET SOCK_DGRAM 0 in
+    let ssock = socket ~cloexec:true PF_INET SOCK_DGRAM 0 in
```
rosalogia, line 48: What does this do?

lib/node/failure_detector.ml
```diff
@@ -1,6 +1,7 @@
 open Common
 open Common.Util
 open Types
+open Bin_prot.Common
```
rosalogia, line 4: Kind of confused; you opened this module but I don't see any changes/additions that make use of opening it?

lib/node/failure_detector.mli
```diff
@@ -1,7 +1,7 @@
 open Types
 open Common
 
-(** Messages sent by the failure detector protocol *)
```
rosalogia, line 4: Keep it as a documentation comment for hovering :smile:

test/address_prop.ml
```diff
@@ -0,0 +1,38 @@
+open QCheck2.Gen
+open Generator.Test
+
+module SUT = Pollinate.Address
```
rosalogia, line 4: Why this name?

test/address_prop.ml
```diff
@@ -0,0 +1,38 @@
...
+let create =
+  QCheck2.Test.make ~count:1000
+    ~name:"create from string and port is successful" (pair string int)
+    (fun (address, port) ->
+      let addr = SUT.create address port in
+      addr.port == port && addr.address == address)
+
+let from_sockaddr =
+  QCheck2.Test.make ~count:1000
+    ~name:"from_sockaddr from any sockaddress is successful" sockaddress_gen
```
rosalogia, line 15: Can we make it more obvious where sockaddress_gen is defined by specifying the module it's from or opening it at the beginning of this function?

test/generator.ml
```diff
@@ -0,0 +1,59 @@
...
+  let address_gen =
+    pair (pure "127.0.0.1") int >|= fun (address, port) ->
+    Address.{ address; port }
+
+  let peer_gen =
+    let open Peer in
+    address_gen >|= fun address ->
+    { address; status = Alive; neighbors = Base.Hashtbl.Poly.create () }
+
+  let peer_print peer =
```
rosalogia, line 31: We can just put this in the peer module so consumers can use it too if they wish

test/generator.ml
```diff
@@ -0,0 +1,59 @@
+open Messages
+
+module type MTest = sig
```
rosalogia, line 3: Do we need this? Can't we just expose everything?

test/messages.ml
```diff
@@ -0,0 +1,22 @@
+(** Utils function shared by the different tests modules *)
```
rosalogia, line 1: I think we could more accurately describe this module

test/peer_prop.ml
```diff
@@ -0,0 +1,51 @@
...
+  QCheck2.Test.make ~count:1000
+    ~name:"add_peer on empty neighbors leads to size of 1"
+    (pair peer_gen peer_gen) (fun (peer, neighbor) ->
+      let _ = SUT.add_neighbor peer neighbor in
+      Base.Hashtbl.length peer.neighbors == 1)
+
+let add_peers =
+  QCheck2.Test.make ~count:1000
+    ~name:
+      "add_peers results in neighors having the same length that given \
```
rosalogia, line 16: This should NOT be true if any of the peers in the generated list have the same address, right?

rosalogia (changes requested):
Good PR, will be happy to have more tests in the codebase.

lib/common/util.ml
gsebil08: Comment got lost on different re-work:
https://github.com/marigold-dev/pollinate/pull/19#discussion_r838552390

lib/node/failure_detector.ml
gsebil08: Removed, was a wrong tentative on working on nix/macos branch.

test/address_prop.ml
gsebil08: `sockaddress_gen` is coming from `Generator.Test` module, which defines all the generators needed on our suite tests.
Do you want me to remove the `open Generator.Test` at the beginning and here use `Generator.Test.sockaddress_gen` instead?
It's ok for me, so as you wish 😄

test/generator.ml
gsebil08: As you want. I found it was overrated to have a dedicated .mli file, but maybe too confusing to don't expose only the `signatures`...
So as you wish, I'm ok for all 😅

test/generator.ml
gsebil08: Of couse, great idea. Also added the `status`.

test/peer_prop.ml
gsebil08: That's an other property, working on a dedicated test for it. Thanks for pointing it out 👍

lib/node/failure_detector.mli
gsebil08: With two stars?

test/messages.ml
gsebil08: Done.

test/peer_prop.ml
gsebil08: New property test added!

rosalogia (approved):
This seems good to me, good work on adding all these property-based tests.

test/address_prop.ml
gsebil08: Done, also renamed to `Generators` to be more consistent with other modules.

lib/common/peer.ml
```diff
@@ -10,6 +10,12 @@ type t = {
   neighbors : (Address.t, t) Base.Hashtbl.t;
 }
 
+let peer_print peer =
```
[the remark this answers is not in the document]
gsebil08: For the moment, I got rid of it. I think it should be re-worked in #10.
It was just a remainder of a previous PBT.

PR #25, opened 2022-04-04
Mutate client's peer list when updating peer state in failure detector

### LINKED ISSUE #24: Failure detector does not change the state of the client's peer list
SUMMARY: The failure detector transforms the client into a peer and then operates on the peer's list of neighbors. However, this does not achieve the intended effect, since actually we want the client's peer list to be mutated to properly represent the state of its neighbors.

IMPACT: The failure detector is pretty much completely dysfunctional without this being fixed.

SOLUTION: Change the failure detector such that
1. We stop changing the client to a peer just so that it has a list of neighbors. This is no longer necessary.
2. Directly mutate the client's peer list in the failure detector.

RELATED PRs:
- [x] #21
- [ ] #25

rosalogia:
Closes #24 by removing any client-to-peer conversions that existed for the sake of having a peer list to operate on, and instead operating directly on the client's peer list.

### DIFF
--- a/lib/client.ml
+++ b/lib/client.ml
@@ -193,14 +193,23 @@ module Failure_detector = struct
       else
         elem :: pick_random_neighbors neighbors (number_of_neighbors - 1)
 
-  let update_neighbor_status peer neighbor status =
+  (** Updates a peer in the client's peer list with
+  the given status. Returns a result that contains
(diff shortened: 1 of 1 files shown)


gsebil08 (approved):
LGTM! Thanks

PR #29, opened 2022-04-05
ci(ocaml5): get rid of ocaml 4.08 and add ocaml 5

gsebil08:
## Depends
- [x] #28


.github/workflows/esy-test.yml
```diff
@@ -0,0 +1,43 @@
+name: Pollinate tests
```
gsebil08, line 1: Added the `esy` version, since we are using `esy x dune test`

rosalogia:
I don't know much about CI, so this looks okay to me, but it's probably best if we have @ulrikstrid weigh in if you're unsure about some things.

.github/workflows/esy-test.yml
```diff
@@ -0,0 +1,42 @@
...
+      - main
+  schedule:
+    # Prime the caches every Monday
+    - cron: 0 1 * * MON
+
+jobs:
+  test:
+    strategy:
+      fail-fast: false
+      matrix:
```
gsebil08, line 16: @ulrikstrid I think it exists a better way to only use one `os`, right?

ulrikstrid: Yeah, just remove the matrix all together, the matrix is only useful if you want to test multiple combinations

.github/workflows/opam-build.yml
```diff
@@ -17,21 +17,19 @@ jobs:
...
 
-      - name: Use OCaml ${{ matrix.ocaml-compiler }}
+      - name: Use OCaml 5
         uses: ocaml/setup-ocaml@v2
-        with:
-          ocaml-compiler: ${{ matrix.ocaml-compiler }}
+
+      - run: opam repo add beta https://github.com/ocaml/ocaml-beta-repository.git
+
+      - run: opam switch reinstall . --no-install --packages ocaml.5.00.0
```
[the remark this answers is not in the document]
ulrikstrid: I can't come up with a better solution so this is probably fine

gsebil08:
@rosalogia @ulrikstrid should I keep the 4.13 also?
I think it is useless since I added the `esy` workflow, which will test the `4.13` since we have it within `package.json`
```json
  "dependencies": {
    "ocaml": "4.13.x"
    [...]
  }
```
What do you think?

ulrikstrid:
@gsebil08 Since the goal is to use this via deku I would rather have ocaml 5 in esy and then test other versions in opam

rosalogia (approved):

PR #30, opened 2022-04-05
Refactor library into sub-libraries

### LINKED ISSUE #26: Failure detector is stuffed into client module
### Summary

Currently, the failure detection module is kind of stuffed into the client module due to a cyclic dependency issue. The failure detector type must be used when defining the type of the client. However, the failure detector includes various functions that require the client as input. This can be solved by defining the failure detector types before the client type, and then defining the failure detector functions after the client type. Currently, we have achieved this by moving the entire failure detector to the client module, and moving failure detector types out of the failure detector module and above the client type. This is messy and makes navigating and finding issues in the code a real pain. If we can find a way to move things into separate files, we should.

### Relevant Documentation and Links

* [Definition of the failure detector types](https://github.com/marigold-dev/pollinate/blob/main/lib/client.ml#L6)
* [Definition of the failure detector module](https://github.com/marigold-dev/pollinate/blob/main/lib/client.ml#L133)
* [Functions in the failure detector module that depend on the client](https://github.com/marigold-dev/pollinate/blob/main/lib/client.ml#L133)
* [Functions in the client that depend on the failure detector module](https://github.com/marigold-dev/pollinate/blob/main/lib/client.ml#L316)

### Related PRs and Issues:
- #21

rosalogia:
# What Did You Do ???

I ...

* Learned that you can split a library up into several libraries that depend on each other (not cyclically though)
  * Learned that while doing this, you can still have top-level exposed values so that if you have a library, say `node` which exposes a module `Node`, you can have sub-modules such as `Node.Server` as well as top-level values such as `Node.init`

After learning this, I decided to:

* Split the entire codebase into two libraries: `common` and `node`.
  * `common` is for things like the `Tqueue` module, the `Address` module, etc. Modules that define types of data and operations on them that must be accessible from any library in Pollinate.
  * `node` is where anything only needed by the former `client` module goes.
    * A `node` is what used to be called a client.
    * It has a client-part and a server-part, and both of these parts have their own module under the `Node` module.
    * The client is responsible for sending and receiving messages to-and-from other nodes.
    * The server is responsible for handling incoming messages asynchronously.
* Separate failure detector back out into its own file, yay :smile:

# Okay... Is Anything Broken?

Nope :smile: this was a successful refactor. Everything still works, tests still pass.

# How Much of a Pain Will This Be to Review?

The nature of review is that you kind of have to comb through everything that changed and make sure that nothing slipped by the author of the PR. I don't want to impose this on you, and since things work and this is just a refactoring, I don't think this needs as serious review as other PRs might. No functionality has changed, so I would say: review this for the sake of understanding/critiquing the new architecture, don't feel obligated to search for small issues that I may have missed. Nothing wrong with making commits to fix those issues in the future. If you find them anyway, thanks! Please let me know, no harm in fixing them now either, but it's just not the priority since we have much more to do.

Closes #26


lib/common/util.ml
```diff
@@ -3,30 +3,30 @@ open Lwt
 let ( let* ) = Option.bind
 
 module Encoding = struct
-  open Bin_prot
   open Bin_prot.Utils
+  module BCommon = Bin_prot.Common
```
gsebil08, line 7: Clever alias!

lib/node/types.ml
```diff
@@ -0,0 +1,26 @@
+open Lwt_unix
+open Common
+
+type failure_detector_config = {
```
gsebil08, line 4: Should it belong to `failure_detector`?

gsebil08 (approved):
LGTM!
Massive re-organization, which i definitively better! Thanks!

lib/node/types.ml
rosalogia: Cyclic dependency issue. The types module exists so that we don't have modules depending on one another 😅

PR #31, opened 2022-04-06
feat(implementSuspicion): node are suspected before being declared as faulty

gsebil08:
### Current behavior

If we don't receive neither direct or indirect ACK from a peer, we flag it as `Faulty`. This is wrong in the case of the SWIM protocol.

### This PR behavior

If the client didn't get neither direct nor indirect ACK, the peer is flagged as `suspicious`

and 

Add a `suspicious_detector` function, which will be run within the `server.run` and do:
- retrieve all `suspicious` peers
- wait for `suspicion_time` to be reached
- retrieve all `suspicious` peers again
- get the intersection of the two previous lists (it means they were suspicious before and still are)
- remove these peers

## Still need to be done

- Disseminate the `peer_n is alive` message to n randomly picker known_peers
- Disseminate the `peer_n is suspicious` message to n randomly picked known_peers
- Disseminate the `peer_n is Faulty` message to n randomly picker known_peers

But I think we can do that in an other dedicated PR once this one is merged and OK. @rosalogia what is your opinion?


lib/node/failure_detector.ml
```diff
@@ -168,3 +170,34 @@ let failure_detection node =
           Lwt_unix.sleep @@ Float.of_int t.config.protocol_period;
         ] in
     Lwt.return ()
+
+module PeerSet = Set.Make (struct
+  let compare = Stdlib.compare
+  type t = Peer.t
+end)
+
+let suspicious_detection node =
```
gsebil08, line 179: @rosalogia I made a basic function and will need help to make it great 😅 
This function is just run within the `server.run` and simply do, totally async of the failure_detector:
- get all suspicious nodes
- wait for timeout to be reached
- get all suspicious nodes
- intersection of the two lists (to retrieve peers which were suspicious before and after timeout)
- remove the peers present in the intersection

rosalogia: Is there a reason the function can't be integrated into the failure detector function?

lib/node/failure_detector.ml
```diff
@@ -62,11 +62,11 @@ let rec pick_random_neighbors neighbors number_of_neighbors =
     else
       elem :: pick_random_neighbors neighbors (number_of_neighbors - 1)
 
-(** Updates a peer in the node's peer list with
```
rosalogia, line 65: Keeping this as a documentation comment means that when you hover over this function's name elsewhere in the code, you can see the comment. I think this is quite useful :smile:.

lib/node/failure_detector.ml
```diff
@@ -168,3 +170,37 @@ let failure_detection node =
           Lwt_unix.sleep @@ Float.of_int t.config.protocol_period;
         ] in
     Lwt.return ()
+
+module PeerSet = Set.Make (struct
+  let compare = Stdlib.compare
```
rosalogia, line 175: How exactly does using `Stdlib.compare` for `Peer.t` work?

lib/node/failure_detector.ml
```diff
@@ -168,3 +170,37 @@ let failure_detection node =
...
+  let open Peer in
+  let t = !node.failure_detector in
+  let suspicious_peers =
+    List.filter
+      (fun p -> p.status = Peer.Suspicious)
+      (Base.Hashtbl.data !node.peers) in
+  match List.length suspicious_peers with
+  | 0 -> Lwt.return ()
+  | _ ->
+    let _ = Lwt_unix.sleep @@ Float.of_int t.config.suspicion_time in
```
rosalogia, line 189: I have no clue if this does anything if we never pass it to bind

lib/node/failure_detector.ml
```diff
@@ -168,3 +170,37 @@ let failure_detection node =
...
+    let suspicious_peers' =
```
rosalogia, line 190: What's the idea behind doing the same filtration a second time here?

lib/node/failure_detector.ml
```diff
@@ -168,3 +170,37 @@ let failure_detection node =
...
+      List.filter
+        (fun p -> p.status = Peer.Suspicious)
+        (Base.Hashtbl.data !node.peers) in
+    let s1 = PeerSet.empty in
+    let s1 = List.fold_right PeerSet.add suspicious_peers s1 in
```
rosalogia, line 195: There is `Set.S.of_list` for this

lib/node/failure_detector.ml
```diff
@@ -168,3 +170,37 @@ let failure_detection node =
...
+    let s2 = PeerSet.empty in
+    let s2 = List.fold_right PeerSet.add suspicious_peers' s2 in
```
rosalogia, line 197: What's the difference? They seem like identical sets

lib/node/failure_detector.ml
gsebil08: Nope. I just thought it would be better to separate them.

rosalogia: Sure :+1: We'll run it right under the current call to failure detector function

lib/node/failure_detector.ml
gsebil08: Good question!
I guess adding some tests would help us to determine this :)

lib/node/failure_detector.ml
gsebil08: Bound it.

lib/node/failure_detector.ml
gsebil08: Totally reworked.

rosalogia (approved):
Code makes sense to me :smile: we'll see if it works soon

lib/node/failure_detector.ml
```diff
@@ -63,17 +63,25 @@ let rec pick_random_neighbors neighbors number_of_neighbors =
...
-unit if the peer is found in the node's list, and
-a string stating that a peer with the given address
-could not be found otherwise. *)
+   the given status. Returns a result that contains
+   unit if the peer is found in the node's list, and
+   a string stating that a peer with the given address
+   could not be found otherwise. *)
 let update_peer_status node peer status =
   let open Peer in
+  let open Unix in
```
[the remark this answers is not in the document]
gsebil08: Agreed, done.

lib/node/failure_detector.ml
gsebil08: Outdated, this `set` does not exist anymore :)

lib/node/failure_detector.ml
```diff
@@ -63,17 +63,24 @@ let rec pick_random_neighbors neighbors number_of_neighbors =
...
+   the given status. Returns a result that contains
+   unit if the peer is found in the node's list, and
+   a string stating that a peer with the given address
+   could not be found otherwise. *)
 let update_peer_status node peer status =
   let open Peer in
   let neighbor = Base.Hashtbl.find !node.peers peer.address in
   match neighbor with
-  | Some neighbor ->
+  | Some neighbor -> (
```
rosalogia, line 74: How about

```ocaml
match neighbor with
| Some neighbor when status = Suspicious ->
  let time = Unix.gmtime @@ Unix.time () in
  Result.Ok(neighbor.last_suspicious_status <- Some time)
| Some neighbor ->
  Result.Ok(neighbor.last_suspicious_status <- None)
| None ->
  Result.Error (Printf.sprintf "Failed to find peer with address %s:%d in node peer list"
         peer.address.address peer.address.port)
```

rosalogia (approved):
Looks good, obviously we don't know yet if it works but from what I can see there are no major issues.

lib/node/failure_detector.ml
gsebil08: Yup! Didn't know this syntax, looks great.
Done.

test/failure_detector_tests.ml
```diff
@@ -0,0 +1,49 @@
...
+let node_b =
+  Lwt_main.run
+    (Node.init ~router:Commons.router ~state:["test1"]
+       ~msg_handler:Commons.msg_handler ("127.0.0.1", 3004))
+
+let peer_b = Client.peer_from !node_b
+
+let suspicion_detection () =
+  let open Common.Peer in
+  let _ = add_neighbor (Pollinate.Node.Client.peer_from !node_a) peer_b in
```
rosalogia, line 20: This works but suspiciously so. It only works because when you construct a peer from a client, the underlying reference to the neighbors/peers hash table is the same. Thus, any change to the neighbors table in the result of `peer_from !node_a` will affect `node_a`'s peer table. I'm not sure we should do this, especially if we will eventually change to using maps which may have different functionality. Instead, let's add a function to the client for adding peers.

test/failure_detector_tests.ml
```diff
@@ -0,0 +1,49 @@
...
+  let _ = SUT.update_peer_status node_a peer_b Suspicious in
+  let%lwt _ = SUT.suspicious_detection node_a in
+  Lwt.return (Base.Hashtbl.length !node_a.peers = 0)
```
rosalogia, line 23: It's nice that we can break away a small bit of functionality and test it this way, but `suspicious_detection` probably will not be called this way in the library, so we should try and later add some integration tests that ensure the suspicion detector is performing as expected when combined with the rest of the library.

rosalogia (approved):
I had one suggestion, but feel free to fix it and merge without asking for review again.

test/failure_detector_tests.ml
gsebil08: Totally agreed. This is clearly irrelevant of the targeted way to use this function. But I wanted to make a quick test about its perimeter, in order to be sure the status are well calculated and the Hashtbl is well updated.
I guess it will be reworked in some future integration tests, but it is still enough right now to ensure this function has a correct behavior. Don't you think?

test/failure_detector_tests.ml
gsebil08: Oh boy! That's a great catch!
I added a `add_peer` function on Client.ml, but should we still propose the `add_neighbor`?

PR #33, opened 2022-04-11
Generalized message dissemination

rosalogia:
This PR implements generalized dissemination of messages in Pollinate. The implementation introduces the notion of "dissemination rounds". Throughout execution of the application server, a pool of messages that should be disseminated across the network is maintained, and each message in the pool is sent to a configurable number of randomly chosen peers at each round for some configurable number of rounds. The disseminator itself is not responsible for message broadcasting, but it maintains the list of messages that need to be broadcasted.


rosalogia:
Current issue is that the message handler has to change so that the consumer can choose how to handle any incoming messages, not just those with category `Request`. Obviously, though, we don't want the message handler to handle incoming _responses to requests_, so we need to find a clean way to divide these two groups of messages.

lib/node/failure_detector.ml
```diff
@@ -79,26 +79,27 @@ let update_peer_status node peer status =
...
 
-let send_message message node (recipient : Peer.t) =
-  let message = create_message node message recipient in
-  Client.send_to node message
+let send_message message node (recipients : Peer.t list) =
+  let message = create_message node message recipients in
+  Networking.send_to node message
 
-let send_ping_to node peer = send_message Ping node peer
+let send_ping_to node peer = send_message Ping node [peer]
```
gsebil08, line 97: Hum, I guess there is still some work to do within the `handle_message` function, where we do:
```ocaml
let _ = List.map (send_ping_request_to node) pingers in
```

lib/node/message.ml
```diff
@@ -5,15 +5,27 @@ type category =
...
   payload : bytes;
 }
 [@@deriving bin_io]
+
+let hash_of m =
+  [
+    m.sender.address;
+    string_of_int m.sender.port;
+    string_of_float m.timestamp;
+    Bytes.to_string m.payload;
```
gsebil08, line 28: Should we add the `id`?
Same question about recipients, but I have a strong feeling that it is not useful.

lib/node/server.ml
```diff
@@ -22,32 +21,80 @@ let handle_response request_table res =
...
-      | Some message -> Failure_detector.handle_message node message
-      | None -> Lwt.return () in
-
-    let%lwt () = Failure_detector.failure_detection node in
-
-    let%lwt next_response = Inbox.next !node.inbox Message.Response in
-    let _ = handle_response !node.request_table next_response in
-
-    let%lwt request = Inbox.next !node.inbox Message.Request in
+let rec run last node preprocessor msg_handler =
```
gsebil08, line 24: I think this function definitively needs some comments inside its scope. I found it pretty hard to read and understand :/

lib/node/disseminator.ml
```diff
@@ -0,0 +1,38 @@
+type pool_elt = {
+  message : Message.t;
+  remaining : int;
+}
+
+module DigestSet = Set.Make(Digest)
+
+type t = {
```
gsebil08, line 8: I feel like making everything `mutable` would be the same as making nothing `mutable`... Wouldn't it be better to return a new `Disseminator.t` instead of having all these mutable types?

lib/node/networking.ml
```diff
@@ -0,0 +1,79 @@
...
+    Encoding.read_size_header size_buffer + Encoding.size_header_length in
+  let msg_buffer = Bytes.create msg_size in
+  (* Now that we have read the header and the message size, we can read the message *)
+  let%lwt _ = recvfrom node_socket msg_buffer 0 msg_size [] in
+  let message = Encoding.unpack Message.bin_read_t msg_buffer in
+  Mutex.unlock !node.socket;
+  Lwt.return message
+
+(** Basic random shuffle, see https://en.wikipedia.org/wiki/Fisher%E2%80%93Yates_shuffle*)
+let knuth_shuffle known_peers =
```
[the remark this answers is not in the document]
rosalogia: Yep

lib/node/networking.ml
```diff
@@ -0,0 +1,83 @@
...
+  recipients
+  |> List.map (fun recipient ->
+         Message.
+           { category; id = -1; sender = !node.address; recipient; payload })
+  |> List.map (fun msg ->
+         let%lwt () = send_to node msg in
+         Lwt.return ())
+
+let disseminate node =
+  let dissemination_group = pick_random_neighbors !node.peers 10 in
```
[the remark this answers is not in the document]
rosalogia: Right, hmm.. Anyway, this should probably be configurable

lib/node/disseminator.ml
rosalogia: Handled

lib/node/server.ml
rosalogia: Yeah, that's definitely true.

lib/node/types.mli
```diff
@@ -59,5 +59,5 @@ type 'a node = {
   (* Dissemination component ; runs automatically with the server and is responsible
      for automatically disseminating both new Post messages and received Post messages
      with other nodes in the network *)
-  disseminator : Disseminator.t;
+  mutable disseminator : Disseminator.t;
```
gsebil08, line 62: What are the advantages to make it mutable?

lib/node/disseminator.ml
```diff
@@ -0,0 +1,53 @@
...
+  num_rounds : int;
+  epoch_length : float;
+  seen : DigestSet.t;
+}
+
+let create num_rounds epoch_length =
+  { round = 0; pool = []; num_rounds; epoch_length; seen = DigestSet.empty }
+
+let next_round disseminator =
+  let time = Unix.time () in
```
gsebil08, line 20: `Unix.time` is in seconds, is it precise enough?

lib/node/disseminator.mli
```diff
@@ -0,0 +1,37 @@
...
+    dissemination component. *)
+type t
+
+(** Creates a dissemination component that can be attached to a node
+    when given the number of "rounds" for which each new message should
+    be disseminated and an "epoch length", in seconds, which determines
+    whether a message is too old to be disseminated again by checking
+    whether the message is newer than n seconds old, where n is the
+    given epoch length. *)
+val create : int -> float -> t
```
gsebil08, line 14: What about creating custom types to make this signature more readable?
Like:
```ocaml
type rounds = int
type epoch_length = float

val create : rounds -> epoch_length -> t
```
I think we have several functions where such custom types would help the readability, what do you think?

lib/node/networking.ml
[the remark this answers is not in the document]
gsebil08: I don't it should be configurable. From my point of view, it is Pollinate responsibility to know how many peers should be in this group.

lib/node/disseminator.mli
rosalogia: We can also just use named parameters, the effect is the same

lib/node/disseminator.ml
rosalogia: I think seconds is precise enough yes, this is just a matter of whether we think a message is too old to be "hot" anymore

lib/node/types.mli
rosalogia: If the fields of the disseminator are not mutable, then the disseminator needs to be mutable. Ultimately because the disseminator maintains state: the messages that are currently being disseminated and how many rounds are left before they're no longer hot enough to disseminate.

lib/node/networking.ml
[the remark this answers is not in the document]
rosalogia: Hmm, I don't think there's a perfect number though, since as you said we don't know the whole network. Check out https://www.serf.io/docs/internals/simulator.html. This is one of the first things I looked at when looking into SWIM and Gossip. From this, I got the idea that these values should probably be configurable (aside from the simulated values, which are determined by real world circumstances). But maybe that's not the case; I'm not really sure.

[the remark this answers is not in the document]
gsebil08: Definitively, thanks for sharing this tool!
Let's keep it hard coded for now, we will make it configurable later, if needed.

lib/node/disseminator.mli
```diff
@@ -0,0 +1,44 @@
...
+val broadcast_queue : t -> Message.t list
+
+(** Determines whether the dissemination component has witnessed a
+    given message before, or whether the message is too old to be
+    retained in the set of seen messages. *)
+val seen : t -> Message.t -> bool
+
+(** Returns the 7 digit hashes of all the messages
+    that the disseminator has seen. *)
+val all_seen : t -> string list
```
gsebil08, line 41: I think this name is not the best one. I supposed `all_seen` to return a bool, as `seen` does.
Maybe something like `get_seen_msgs_hash` would be better.

lib/node/client.ml
```diff
@@ -26,8 +24,9 @@ let create_request node recipient payload =
             category = Message.Request;
             sub_category_opt = None;
             id = !id;
+            timestamp = Unix.gettimeofday ();
             sender = !node.address;
-            recipient;
+            recipients = [recipient];
```
gsebil08, line 29: From my point of view, this is not a good change.
In Deku side _(@aguillon and @AngryStitch please correct me if I am wrong)_ we want to either 
- send a Request (to a specific node, and we care about the result)
- broadcast an information (to the whole network, and we don't care about the result)
This is what I understand from `network.ml` file on main branch.

Hence, in Pollinate, we should provide these two features, which are totally different.
1. Sending request/response
2. Broadcasting message (Post in this PR).

All that to say, since the recipient is contained in the message, `create_request` and `create_post` should not return the same kind of Message.
I think I would have use GADT in Haskell for this, but don't know about OCaml :/

gsebil08, line 29: From my point of view, this is not a good change.
In Deku side _(@aguillon and @AngryStitch please correct me if I am wrong)_ we want to either 
- send a Request (to a specific node, and we care about the result)
- broadcast an information (to the whole network, and we don't care about the result)
This is what I understand from `network.ml` file on main branch.

Hence, in Pollinate, we should provide these two features, which are totally different.
1. Sending request/response
2. Broadcasting message (Post in this PR).

All that to say, since the recipient is contained in the message, `create_request` and `create_post` should not return the same kind of Message.
I think I would have use GADT in Haskell for this, but don't know about OCaml :/

lib/node/failure_detector.ml
```diff
@@ -82,27 +82,28 @@ let update_peer_status node peer status =
       (Printf.sprintf "Failed to find peer with address %s:%d in node peer list"
          peer.address.address peer.address.port)
 
-let create_message node message (recipient : Peer.t) =
+let create_message node message (recipients : Peer.t list) =
```
gsebil08, line 85: In SWIM paper, a message is sent to only one recipient.
Only the messages about node state (for example "a peer is suspicious") are disseminated. The others, are kinda request/response via direct messaging.
This is kinda related to my other comment on Client.ml

gsebil08, line 85: Hum, I am not sure about this change.
In SWIM paper, a message is sent to only one recipient.
Only the messages about node state (for example "a peer is suspicious") are disseminated. The others, are kinda request/response via direct messaging.

lib/node/failure_detector.ml
```diff
@@ -82,27 +82,28 @@ let update_peer_status node peer status =
...
+let send_ping_to node peer = send_message Ping node [peer]
 
-let send_acknowledge_to node peer = send_message Acknowledge node peer
+let send_acknowledge_to node peer = send_message Acknowledge node [peer]
 
 let send_ping_request_to node (recipient : Peer.t) =
-  send_message (PingRequest recipient.address) node recipient
+  send_message (PingRequest recipient.address) node [recipient]
 
 let handle_message node message =
```
gsebil08, line 108: Function knuth_shuffle from this file should be removed, it is defined in `Networking`.

gsebil08, line 108: Function knuth_shuffle from here should be removed, it is defined in `Networking`.

lib/node/client.mli
```diff
@@ -40,3 +36,5 @@ val request : node ref -> bytes -> Address.t -> Message.t Lwt.t
 of recipients and collects the responses in a list of [Message.t Lwt.t]. *)
 val broadcast_request :
```
gsebil08, line 37: This one should be either renamed or deleted.
If we want to broadcast a message, we should use the dissemination. If we want to send direct message to everyone _(can't see any use case where such a feature would be useful)_, we should rename it.

gsebil08, line 37: This one should be either renamed or deleted.
If we want to broadcast a message, we should use the dissemination. If we want to send direct message to everyone _(can't see any use case where such a feature would be useful)_, we should rename it.

lib/node/server.ml
```diff
@@ -1,51 +1,125 @@
...
+    | Failure_detection -> Failure_detector.handle_message node message
+    | Post ->
+      if not (Disseminator.seen !node.disseminator message) then (
+        let%lwt () =
+          log node
+            (Printf.sprintf "%s:%d : Processing post %s from %s:%d\n"
+               !node.address.address !node.address.port
+               (Message.hash_of message) message.sender.address
+               message.sender.port) in
+        let _ = msg_handler message in
```
gsebil08, line 61: This line seems totally useless, right?

gsebil08, line 61: This line seems totally useless, right?

lib/node/server.ml
```diff
@@ -1,51 +1,125 @@
...
+        let%lwt () = log node "Adding message to broadcast queue\n" in
+        Client.post node message;
+        Lwt.return ())
+      else
+        log node
+          (Printf.sprintf "Got post %s from %s:%d but saw it already\n"
+             (Message.hash_of message) message.sender.address
+             message.sender.port)
+    | _ ->
+      let _ = msg_handler message in
```
gsebil08, line 71: Same here, useless, right?

gsebil08, line 71: Same here, useless, right?

lib/node/server.ml
```diff
@@ -1,51 +1,125 @@
...
+  (* Step 0 *)
+  let%lwt () = print_logs node in
+  (* Step 1 *)
+  let _ = process_message node preprocessor msg_handler in
+  (* Step 2 *)
+  let%lwt () = Failure_detector.failure_detection node in
+  (* Step 3 *)
+  let%lwt () = Networking.disseminate node in
+  (* Step 4 *)
+  let%lwt () = Lwt_unix.sleep 0.001 in
```
gsebil08, line 124: Hum, not fond of this, but can't find an other way to achieve this.
Aren't the execution of `failure_detection` and `disseminate` enough to "lose" time?

gsebil08, line 124: Hum, not fond of this, but can't find an other way to achieve this.
Aren't the execution of `failure_detection` and `disseminate` enough to "lose" time?

lib/node/server.mli
```diff
@@ -0,0 +1,8 @@
+(** Runs the server given a reference to a node, a message preprocessor and a message handler.
+    The server is responsible for running failure detection and dissemination processes, as well
+    as issuing responses to nodes making individual requests via the message handler. *)
+val run :
+  Types.node ref ->
+  (Message.t -> Message.t) ->
+  (Message.t -> bytes option) ->
```
gsebil08, line 7: This is a great change!

gsebil08 (changes requested):
This is a massive work you have done!
Thanks for handling this 🙏

lib/node/disseminator.mli
rosalogia: Good point, thanks 👍

lib/node/server.ml
rosalogia: I can fix this later with Lwt.async

gsebil08 (approved):

lib/node/client.ml
gsebil08: Let's open a dedicated issue for that. This is absolutely not a blocker.

lib/node/message.ml
```diff
@@ -5,16 +5,30 @@ type category =
   | Uncategorized
   | Request
   | Response
+  | Post
```
gsebil08, line 8: Following Deku integration, this category is totally internal.
Since we only expose `create_request`, `create_response` and `create_post` some categories are never going to be used (`Uncategorized` and `Custom`).

How about getting rid of these useless types, and rename the `sub_category_opt` in the other branch to let the client set it?

rosalogia: It's used internally to know how to process messages though? What exactly is your suggestion?

rosalogia: Sorry to only say this after you've gone on vacation, but I'm not sure I really follow this design/interface; hopefully we can have a good discussion about it when you're back, or if Maria has the same idea I can talk to her about it :smile:.

PR #34, opened 2022-04-13
WIP feat(forwardMessage): Node forwards message if it is not the recipient

Gau-thier:


lib/node/server.ml
```diff
@@ -35,18 +35,29 @@ let run node router msg_handler =
     let%lwt () = Failure_detector.failure_detection node in
 
     let%lwt next_response = Inbox.next !node.inbox Message.Response in
-    let _ = handle_response !node.request_table next_response in
+    let%lwt _ =
```
gsebil08, line 38: The more I think, the more I am lost...
My initial thought was:
> It looks like we handle **every** messages, even if we are not the actual recipient

That's why I added a naive
> I am not the recipient, I just resend the message

But I now see two troubles:
1. This design will send the message in every case. Hence, even if it has already been handled by the actual recipient ==> This makes me think we have to check the sequence_number of the message (also in the failure_detector)
2. This looks so bad for the performance... If every peer does this, we will send so many useless messages...

Should I revert this, and only add a new function, something like:
```ocaml
indirect_send_to : 'a node ref -> Peer.t -> Message.t -> unit Lwt.t
```

PR #35, opened 2022-04-14
feat(broadcastNodeStatus): send node status to everyone

Gau-thier:
Following https://github.com/marigold-dev/pollinate/pull/31 "still need to be done"


rosalogia (approved):

rosalogia:
This looks good, but as mentioned we need to adapt it to dissemination, so not merge-ready

PR #37, opened 2022-04-25
doc(odoc): init odoc

gsebil08:
The changes are directly within the `*.mli` files.
I removed the generated `*.html` to our git repo, @rosalogia should I get rid of them and only add something like
> To generate the doc, run:
> ```shell script
> $ dune build @doc
> ```

in our Readme.md?


rosalogia (approved):

ulrikstrid:
I would strongly advice against adding the *.html files to the repo since they will change often and just add noise. You can setup publishing of docs to github pages instead.

ulrikstrid (approved):
Looks good from infra point of view

PR #40, opened 2022-05-10
dep(removeUseless): remove useless dependency

gsebil08:


gsebil08:
Closing for now, since it works with nix.

PR #41, opened 2022-05-12
Refactor nix and make pollinate consumable

ulrikstrid:
Refactored nix to follow approximately what I did in `json-logs-formatter` which exposes a overlay with pollinate and exposes pollinate under packages for each OCaml version that we support.

Also added `public_name` to main `pollinate` library and exposed the `Inbox` module in `Node`.


flake.nix
```diff
@@ -14,31 +14,63 @@
...
+            });
+          };
 
-          defaultPackage = myPkgs.pollinate;
+          packages = pkgs.lib.mapAttrs (version: ocamlPackages:
+            (pkgs.callPackage ./nix {
+              inherit ocamlPackages;
+              doCheck = true;
+            })) (builtins.removeAttrs pkgs.ocaml-ng [
+              "overrideDerivation"
```
gsebil08, line 42: Does it mean "we want to remove everything except ocaml 4.13"?
Isn't there any more beautiful way to do it?

lib/node/node.ml
```diff
@@ -4,6 +4,7 @@ open Types
 module Message = Message
 module Client = Client
 module Failure_detector = Failure_detector
+module Inbox = Inbox
```
gsebil08, line 7: Why?

nix/overlay.nix
```diff
@@ -0,0 +1,27 @@
+final: prev: {
+  ocaml-ng = (prev.lib.mapAttrs (_: ocamlVersion:
+    ocamlVersion.overrideScope' (oself: osuper: {
+      pollinate = (final.callPackage ./default.nix {
+        ocamlPackages = osuper;
+        doCheck = true;
+      });
+    })) (builtins.removeAttrs prev.ocaml-ng [
```
gsebil08, line 8: Looks like it is exactly the same than in `flake.nix`, can we make it using the same to avoid duplication?

gsebil08 (commented):
Thanks for this, it looks great.
I am wondering if we should only make it OCaml 5.00 compatible... But not sure...

lib/node/node.ml
ulrikstrid: It's used in the deku integration branch i tested against, of this is wrong we should revert and fix downstream

nix/overlay.nix
ulrikstrid: Yes, but I was lazy

flake.nix
ulrikstrid: Yes, we should invert it

ulrikstrid:
If we only want ocaml 5 we can simplify this quite a bit

gsebil08 (approved):

PR #44, opened 2022-05-18
chore(removeState): state is not relevant is blockchain node exchanges

gsebil08:


lib/node/node.ml
```diff
@@ -6,9 +6,7 @@ module Client = Client
 module Failure_detector = Failure_detector
 module Inbox = Inbox
 
-type 'a t = 'a Types.node
-
-let init ~state ?(router = fun m -> m) ~msg_handler ?(init_peers = [])
+let init ?(router = fun m -> m) ~msg_handler ?(init_peers = [])
```
gsebil08, line 9: @rosalogia I first tried to switch to `?state` but can't figure out how to make it works.
Any help will be welcomed 🙏

rosalogia (approved):

PR #45, opened 2022-05-19
feat(msgHandlerIsOptional): switch msg_handler to optional to match real life examples

gsebil08:


gsebil08:
Totally wrong, drop it.

PR #47, opened 2022-06-02
chore(updateNixFalke): run nix flake update to fix CI

gsebil08:
# Context

Nix GitHub action was broken

### DIFF
--- a/flake.nix
+++ b/flake.nix
@@ -4,19 +4,21 @@
     flake-utils.url = "github:numtide/flake-utils";
     nix-filter.url = "github:numtide/nix-filter";
 
-    ocaml-overlay.url = "github:anmonteiro/nix-overlays";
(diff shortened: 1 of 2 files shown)


flake.nix
```diff
@@ -4,19 +4,21 @@
     flake-utils.url = "github:numtide/flake-utils";
     nix-filter.url = "github:numtide/nix-filter";
 
-    ocaml-overlay.url = "github:anmonteiro/nix-overlays";
-    ocaml-overlay.inputs.nixpkgs.follows = "nixpkgs";
-    ocaml-overlay.inputs.flake-utils.follows = "flake-utils";
+    ocaml-overlays.url = "github:anmonteiro/nix-overlays";
```
gsebil08, line 7: Renaming from `overlay` to `overlays` and use the same syntax as Deku to be conssitent.

Issue #51, opened 2022-06-03
Unnecessary sleep time in server


rosalogia:
Currently we require a 0.001 sleep time between server iterations due to the fact that we handle message reception/processing in a secondary thread. This means that the recursive function that acts as the server has nothing to wait for between runs, and thusly sort of spins out of control if there is no sleep time between recursive calls. This is not necessarily the correct design. The reason we handle message reception/processing in a secondary thread is because the server is responsible for triggering some other periodic behaviours such as dissemination and failure detection, and these must fire regularly even if a message is not received. The real solution is to make things like dissemination and failure detection asynchronous with `Lwt.async`, have the server wait for new messages, and handle message processing in a secondary thread.

RELATED PRs:
- [ ] #33

PR #52, opened 2022-06-10
Move background processes to an Lwt.async context

rosalogia:
This PR tries to address #51 by moving certain "background processes" such as sending messages for failure detection and dissemination out of the server and into an `Lwt.async` call. This means the server can block as much as it wants waiting for new messages and handling them before continuing. The original reason for this was to remove the fixed interval between server iterations due to the fact that messages were being received/handled in a separate thread which was not being awaited. However, I have found that this approach does not actually allow us to remove the fixed interval. It is still necessary for the interval to be there in the function called with `Lwt.async`, or things begin to stall. I am not sure if this change is desirable.


gsebil08 (approved):
This is a good change!
I accept, but tests seems to fail, once they are fixed, this can be merged.

PR #56, opened 2022-07-04
Re-add deku needs

Gau-thier:


lib/node/client.ml
```diff
@@ -21,7 +21,7 @@ let add_peer_as_is node (peer : Peer.t) =
 
 let peers node = Base.Hashtbl.keys node.peers
 
-let create_request node ?(request_ack = false) recipient payload =
+let create_request node ?(request_ack = false) recipient (payload, payload_signature) =
```
rosalogia: Since `payload_signature` is represented as an option, why not make it an optional argument?
```ocaml
let create_request node ?(request_ack = false)?payload_signature recipient payload =
```

lib/node/server.ml
```diff
@@ -128,7 +131,7 @@ let _print_logs node =
    3. Run the disseminator, this includes actually sending messages to be
       disseminated across the network.
    4. Wait 0.001 seconds before restarting the procedure. *)
-let rec run node preprocessor msg_handler =
+let rec run node preprocessor (msg_handler: Message.t -> bytes option * bytes option ) =
```
rosalogia: Shouldn't it be `(bytes * bytes option) option`? Since you can't have a payload_signature without a payload?

rosalogia (changes requested):
I think payload_signature should be an optional arg, otherwise looks good :+1:

lib/node/server.ml
Gau-thier: Making `msg_handler: Message.t -> bytes * bytes option` will force the client to send a response to every received message, which is not wanted.
@AngryStitch am I right?

lib/node/client.ml
Gau-thier: Nice remark, done! 😄

lib/pnode/message.ml
```diff
@@ -20,6 +20,7 @@ type t = {
   sender : Address.t;
   recipients : Address.t list;
   payload : bytes;
+  payload_signature_opt : bytes option;
```
rosalogia: No need for opt in the name since the type is clear about this IMO.

lib/pnode/node.mli
```diff
@@ -14,7 +14,7 @@ val init : ?init_peers:Address.t list -> Address.t -> t ref Lwt.t
 
 val run_server :
   ?preprocessor:(Message.t -> Message.t) ->
-  msg_handler:(Message.t -> bytes option) ->
+  msg_handler:(Message.t -> bytes * bytes option) ->
```
rosalogia: Should be `Message.t -> (bytes * bytes option) option` which is kind of visually confusing so maybe we should make a record type?

lib/pnode/server.ml
```diff
@@ -33,7 +33,8 @@ let handle_ack node msg =
 
    Otherwise, we just apply the message handler and that's it.
 *)
-let process_message node preprocessor msg_handler =
+let process_message node preprocessor
+    (msg_handler : Message.t -> bytes * bytes option) =
```
rosalogia: Same as above

test/disseminator_tests.ml
```diff
@@ -13,23 +13,23 @@ module Disseminator_tests = struct
...
-    |> string_of_int
-    |> String.to_bytes
-    |> Client.create_post node
-    |> Client.post node;
+    let payload =
+      Client.address_of !node
+      |> (fun Address.{ port; _ } -> port)
+      |> string_of_int
+      |> String.to_bytes in
+    Client.create_post node payload ?payload_signature:None |> Client.post node;
```
rosalogia: No need for specifying None here, it's an optional argument so not specifying it at all will default to None

test/disseminator_tests.ml
```diff
@@ -39,12 +39,12 @@ module Disseminator_tests = struct
...
     let _server = Node.run_server ~msg_handler:Commons.msg_handler node in
-    let message =
+    let payload =
       Client.address_of !node
       |> (fun Address.{ port; _ } -> port)
       |> string_of_int
-      |> String.to_bytes
-      |> Client.create_post node in
+      |> String.to_bytes in
+    let message = Client.create_post node payload ?payload_signature:None in
```
rosalogia: Same as above

rosalogia (changes requested):

lib/pnode/message.ml
Gau-thier: You're right, done.

test/disseminator_tests.ml
Gau-thier: Done.

test/disseminator_tests.ml
Gau-thier: Done.

lib/pnode/node.mli
Gau-thier: Done, also re-used it into `Message.t`

lib/pnode/server.ml
Gau-thier: Done.

lib/node/server.ml
AngryStitch: Shouldn't it be `msg_handler: Message.t -> (bytes * bytes) option` because if you have a payload, you have a signature but sometimes you don't do anything with a message (like, it's an ACK)?

Gau-thier: On a Deku POV, "if you have a payload you have a message" but not on a Pollinate side.
Failure_detector is sending payload message, without signature.
Maybe we should propose to add a function to sign them, to make sure the node sending is the real node, or is it overkilled? @rosalogia, what is your opinion?

lib/pnode/server.ml
```diff
@@ -52,13 +53,15 @@ let process_message node preprocessor msg_handler =
               !node.address.address !node.address.port message.sender.address
               message.sender.port) in *)
       match msg_handler message with
-      | Some response ->
-        response
-        |> Client.create_response node message
+      | Some msg ->
```
rosalogia: We can destructure here (just a matter of syntax); up to you
```ocaml
| Some Message.{ data ; signature = payload_signature }
  Client.create_response node message ?payload_signature data

File: test/disseminator_tests.ml [hunk 9]
```diff
@@ -1,53 +1,53 @@
...
-    let message =
+    let _server = Pnode.run_server ~msg_handler:Commons.msg_handler node in
+    let payload =
       Client.address_of !node
       |> (fun Address.{ port; _ } -> port)
       |> string_of_int
-      |> String.to_bytes
-      |> Client.create_post node in
+      |> String.to_bytes in
+    let message = Client.create_post node payload in
```
rosalogia, line 47: I will always advocate for long pipelines, but ultimately doesn't matter :smile:

rosalogia (approved):

test/disseminator_tests.ml
Gau-thier: I am still not fond of long pipelines, I will let as it is right now 😉

lib/pnode/server.ml
Gau-thier: Nice one, totally missed it. Done.

rosalogia:
Forgot to bring this up in review: what is the purpose of sub-category? Why is it not just encoded in the payload? If we want a general purpose way to allow users to include their own metadata/header data in messages, we can do that, but this feels oddly specific for no particular reason. It is always possible for the consumer to design their payloads to include any kind of labels necessary on their end. If we want to do pollinate-side metadata for consumer use, I'd say add a field called "headers" to Message.t whose type is (string * string) list. Thoughts @gsebil08?

Gau-thier:
> Forgot to bring this up in review: what is the purpose of sub-category? Why is it not just encoded in the payload? If we want a general purpose way to allow users to include their own metadata/header data in messages, we can do that, but this feels oddly specific for no particular reason. It is always possible for the consumer to design their payloads to include any kind of labels necessary on their end. If we want to do pollinate-side metadata for consumer use, I'd say add a field called "headers" to Message.t whose type is (string * string) list. Thoughts @gsebil08?

That is a great remark!
It totally makes sens. I just implemented it! Thanks!

PR #57, opened 2022-07-05
Chore/deku formatting

Gau-thier:


flake.nix
```diff
@@ -27,23 +27,23 @@
...
               buildInputs = with pkgs;
                 with ocamlPackages_dev; [
                   ocaml-lsp
-                  ocamlformat
+                  ocamlformat_0_20_1
                   odoc
                   ocaml
                   dune_3
-                  nixfmt
+                  treefmt
```
ulrikstrid, line 43: You probably still want `nixfmt` since `treefmt` will call out to it

nix/formatter.nix
```diff
@@ -0,0 +1,14 @@
...
+
+pkgs.writeShellApplication {
+  name = "format.sh";
+  runtimeInputs = with pkgs; [
+    treefmt
+    nixfmt
+    ocamlformat_0_20_1
+    ocaml
+    dune_3
+    nodePackages.prettier
```
ulrikstrid: Prettier can be removed

ulrikstrid (changes requested):
Added some comments about the dependencies, but looks good otherwise

flake.nix
Gau-thier: My bad, done.

nix/formatter.nix
Gau-thier: Thanks, forgot to remove it. Done.

ulrikstrid (approved):
Looks good

rosalogia (approved):

PR #59, opened 2022-07-07
Added broadcast function for posting without a server

rosalogia:


lib/node/client.ml
```diff
@@ -85,3 +85,6 @@ let request node request recipient =
 
 let post node message =
   !node.disseminator <- Disseminator.post !node.disseminator message
+
+let broadcast node message =
```
Gau-thier, line 89: Already added it in #56 , using `Networking.broadcast` that was already there.

